PR #107, opened 2016-01-29
CRAFTERCMS-1743 - Make Preview Content puck a panel

vestrada:


russdanner:
Please do not comment out code.  Remove it. That's why we use source control :)

Remove commented out code and add to PR

PR #118, opened 2016-02-01
CRAFTERCMS-1735

dejan-brkic:
fixed issue


src/main/java/org/craftercms/studio/impl/v1/service/deployment/PublishingManagerImpl.java
```diff
@@ -407,11 +407,17 @@ public void processItem(CopyToEnvironment item) throws DeploymentException {
...
+                        if (contentService.contentExists(fullPath)) {
+                            try {
+                                RepositoryItem[] children = contentRepository.getContentChildren(fullPath);
+                                if (children.length < 2) {
+                                    deployer.deleteFile(item.getSite(), item.getOldPath().replace("/" + DmConstants.INDEX_FILE, ""));
+                                } else {
+                                    hasRenamedChildren = true;
+                                }
+                            } catch (Exception exc) {
+                                LOGGER.info("Error while checking children for moved content site " + item.getSite() + " old path " + item.getOldPath());
```
russdanner: Exception prints out info.  Should this be an Error instead?  I think so

Also, it swallows the error... so we need to at least print out the error.

russdanner: Please fix and commit.  I will then approve the PR

PR #120, opened 2016-02-02
CRAFTERCMS-1760: added a site component datasource

hlim:


russdanner:
I will merge but 2 thing: 
#1: error messages need to go in message bundles
#2: Would like to get rid of alert boxes.  Need to talk to vero about a common way to spin those up.

PR #125, opened 2016-02-09
Remove commented code

dejan-brkic:
Remove commented code


src/main/java/org/craftercms/studio/impl/v1/service/clipboard/ClipboardServiceImpl.java
```diff
@@ -127,7 +120,6 @@ protected String _duplicate(String site, String path, boolean toDraft, String so
                 throw new ContentNotFoundException(path + " does not exist.");
             }
         }finally{
```
russdanner: Empty Finally?

PR #128, opened 2016-02-10
new linked  form controls

Lpereira00:
Should I add this to the blueprints so they can be listed within the default controlers??


russdanner:
Please externalize the strings in this control.  Studio is now internationalized.  Only need to provide english and Spanish translations.  Any questions on this can be asked of Vero.

Make this change and re-submit PR.

I will approve this for now to help but I do not want NON-resources strings in the code.

luis-rivet:
Sure I'll do it asap

PR #138, opened 2016-02-15
craftercms1778

dejan-brkic:
fixed issue


russdanner:
How does this fix the issue? The same error is thrown if document is null later in the code.  If file does not exist, the document = null and the error will fire.

src/main/java/org/craftercms/studio/impl/v1/service/configuration/ContentTypesConfigImpl.java
```diff
@@ -90,7 +90,9 @@ public ContentTypeConfigTO loadConfiguration(String site, String contentType) {
         String configFileFullPath = siteConfigPath + "/" + configFileName;
         Document document = null;
         try {
-            document = contentService.getContentAsDocument(configFileFullPath);
+            if (contentService.contentExists(configFileFullPath)) {
```
russdanner: This check leaves document null which fires same error below.

russdanner:
Need clarification on this

PR #139, opened 2016-02-16
 FIX-ME: Strangely if there's only one widget in the rte config

Lpereira00:
Widgets will not be an array of widgets but an object instead (now taking one or more options)


russdanner:
Reviewed, RD

PR #144, opened 2016-02-18
don't explode when message format fails.

russdanner:


russdanner:
fox in the hen house, reviewed my own code.

PR #150, opened 2016-02-22
New Notification Service

cortiz:
To enable add 
notificationService.v2.enable=true in the properties files and add the require config in the cstudio/config site named notification-config-v2-.xml


russdanner:
Why have all these methods NOTIFY ON THIS, NOTIFY ON THAT....
Why not just send message.  
Im ok if these are helper methods but I don't understand why the first implementation baked the workflow in to the API (so brittle and limiting) and then we did the same thing again.  Also there needs to be a generic method (DID I MISS IT?) so that things like the expiration service can send notificaitions

PR #160, opened 2016-02-23
several issues

dejan-brkic:
Issues related to content shown as deleted after creation

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/job/EmailMessageSender.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/job/EmailMessageSender.java
@@ -103,8 +103,7 @@ public class EmailMessageSender implements Runnable {
 
 			public void prepare(MimeMessage mimeMessage) throws Exception {
 
-				mimeMessage.setRecipient(Message.RecipientType.TO,
-						new InternetAddress(userEmailAddress));
+				mimeMessage.addRecipients(Message.RecipientType.TO,InternetAddress.parse(userEmailAddress));
 				InternetAddress[] replyTos= new InternetAddress[1];
 				if( (replyTo != null) && (!"".equals(replyTo)) )
 				{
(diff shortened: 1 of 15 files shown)


src/main/java/org/craftercms/studio/impl/v1/repository/alfresco/AlfrescoContentRepository.java
```diff
@@ -693,7 +698,7 @@ protected boolean writeContentCMIS(String fullPath, InputStream content) {
                     // if not, create the folder first
                     boolean created = createMissingFoldersCMIS(folderPath);
                     if (created) {
-                        session.clear();
+                        //session.clear();
```
russdanner: Please do not comment out code. just delete it.

src/main/java/org/craftercms/studio/impl/v1/repository/alfresco/AlfrescoContentRepository.java
```diff
@@ -1069,6 +1099,11 @@ protected Session getCMISSession(boolean alfrescoCMIS) {
         Repository repository = repositories.get(0);
         parameter.put(SessionParameter.REPOSITORY_ID, repository.getId());
         Session session = sessionFactory.createSession(parameter);
+        //session.getDefaultContext().setCacheEnabled(false);
```
russdanner: Please dont comment out code. just delete it.  if it needs to be here for some reason there needs to be a comment why

src/main/webapp/default-site/scripts/rest/api/1/content/write-content.post.groovy
```diff
@@ -113,4 +113,5 @@ else {
         }
     }
 }
+sleep(1000);
```
russdanner: Still need this?

russdanner:
Please make updates and re-submit.

PR #161, opened 2016-02-23
minor update

dejan-brkic:
minor update

### DIFF
--- a/src/main/webapp/default-site/scripts/rest/api/1/content/write-content.post.groovy
+++ b/src/main/webapp/default-site/scripts/rest/api/1/content/write-content.post.groovy
@@ -113,5 +113,4 @@ else {
         }
     }
 }
-sleep(1000);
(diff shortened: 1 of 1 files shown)


src/main/java/org/craftercms/studio/impl/v1/repository/alfresco/AlfrescoContentRepository.java
```diff
@@ -1069,6 +1098,11 @@ protected Session getCMISSession(boolean alfrescoCMIS) {
         Repository repository = repositories.get(0);
         parameter.put(SessionParameter.REPOSITORY_ID, repository.getId());
         Session session = sessionFactory.createSession(parameter);
+        //session.getDefaultContext().setCacheEnabled(false);
```
russdanner: Commented out code.

PR #163, opened 2016-02-24
Move, Approve Notification when its inserted into DB instead when its…

cortiz:
… process

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/deployment/DeploymentServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/deployment/DeploymentServiceImpl.java
@@ -114,7 +114,7 @@ public class DeploymentServiceImpl implements DeploymentService {
         try {
             sendContentApprovalEmail(items, scheduleDateNow);
         }catch(Exception errNotify) {
-            logger.error("Error sending approval notification for");
+            logger.error("Error sending approval notification ",errNotify);
         }
 }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/craftercms/studio/impl/v1/service/deployment/DmPublishServiceImpl.java
```diff
@@ -78,7 +79,7 @@ public void publish(final String site, List<String> paths, Date launchDate,
 
         try {
             deploymentService.deploy(site, mcpContext.getPublishingChannelGroup(), pathsToPublish, ld, approver,
-                        mcpContext.getSubmissionComment());
+                        mcpContext.getSubmissionComment(),scheduledDateIsNow );
         } catch (DeploymentException e) {
             logger.error("Error while submitting paths to publish");
```
russdanner: where is my error?

src/main/java/org/craftercms/studio/impl/v1/service/deployment/DeploymentServiceImpl.java
```diff
@@ -110,6 +110,42 @@ public void deploy(String site, String environment, List<String> paths, Date sch
         for (CopyToEnvironment item : items) {
             copyToEnvironmentMapper.insertItemForDeployment(item);
         }
+        // We need to pick up this on Inserting , not on execution!
+        try {
+            sendContentApprovalEmail(items, scheduleDateNow);
+        }catch(Exception errNotify) {
+            logger.error("Error sending approval notification for");
```
russdanner: Where is the exception in the error message?

PR #179, opened 2016-02-29
CRAFTERCMS-1806 - Canned text for rejection reasons

vestrada:
Validating publishOptionComment on approve.js


src/main/webapp/default-site/static-assets/components/cstudio-dialogs/reject.js
```diff
@@ -21,11 +21,11 @@ REJECT_DIALOG_TEMPLATE = ['<div class=\"bd\">'+
 		'<div class="field">'+
 			'<select id="rejectReasonDropDown" class="rejectReasonDropDown">'+
 				 '<option>Select a Reason</option>'+
```
russdanner: Dont these options come from the service call?  They should not be hard-coded

russdanner: Looks like a default to me that gets replaced.  Am I right?

PR #181, opened 2016-03-01
CRAFTERCMS-1813 - adding expand collapse for each widget/panel on dashboard

vestrada:


russdanner:
functionality is not to spec

vestrada:
But What about: CRAFTERCMS-1821 and CRAFTERCMS-1823?

russdanner:
It just how Git works.  Resubmit pull Request when the other is ready.

/R

PR #235, opened 2016-04-13
CRAFTERCMS-1899

dejan-brkic:
fixed issue


russdanner:
Hi Dejan,

Before I merge this just wanted to get feedback on why the general lock service is better than conventional java locking?

/r

russdanner:
BUMP, Dejan see question

dejan-brkic:
General lock service is conventional java locking (using lock objects not synchronized calls) wrapped in our service class (not to repeat same code everywhere, instead we are just using one method call)

dejan-brkic:
Fixed issue with securing urls for services

PR #268, opened 2016-05-31
Firefox - request publish not working, page reloads before request

jvega190:


russdanner:
Use events, not window reloads

PR #304, opened 2016-07-25
CRAFTERCMS-2009

dejan-brkic:
fixed issue


src/main/java/org/craftercms/studio/impl/v1/service/clipboard/ClipboardServiceImpl.java
```diff
@@ -508,7 +508,7 @@ protected void copyContent(String site, String user, DmPasteItemTO pasteItem, St
 
             String nodePath = fullPath;
             if (contentItem.isFolder()) {
-                copyChildren(site, user, path, contentItem, destination + contentItem.getName(), pasteItem, deep, true, copiedItems);
+                copyChildren(site, user, path, contentItem, destination + "/" + contentItem.getName(), pasteItem, deep, true, copiedItems);
```
russdanner, line 511: do we need a check here for // ?

PR #312, opened 2016-08-09
CRAFTERCMS-1700

dejan-brkic:
fixed issue


src/main/java/org/craftercms/studio/impl/v1/service/configuration/SiteEnvironmentConfigImpl.java
```diff
@@ -220,6 +218,9 @@ protected EnvironmentConfigTO loadConfiguration(String key) {
                 for (Element channel : channels) {
                     PublishingChannelConfigTO pcConfigTO = new PublishingChannelConfigTO();
                     pcConfigTO.setName(channel.getText());
+                    if (!checkEndpointConfigured(key, pcConfigTO.getName())) {
+                        logger.warn("Deployment endpoint \"" + pcConfigTO.getName() + "\" is not configured for site " + key);
```
russdanner, line 222: This should be an error

PR #328, opened 2016-08-31
Several issues fixed

dejan-brkic:
CRAFTERCMS-2119;CRAFTERCMS-1740;CRAFTERCMS-2030

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/repository/alfresco/AlfrescoContentRepository.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/repository/alfresco/AlfrescoContentRepository.java
@@ -59,10 +59,7 @@ import org.apache.chemistry.opencmis.commons.enums.AclPropagation;
 import org.apache.chemistry.opencmis.commons.enums.BaseTypeId;
 import org.apache.chemistry.opencmis.commons.enums.BindingType;
 import org.apache.chemistry.opencmis.commons.enums.VersioningState;
-import org.apache.chemistry.opencmis.commons.exceptions.CmisBaseException;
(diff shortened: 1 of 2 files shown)


src/main/java/org/craftercms/studio/impl/v1/repository/alfresco/AlfrescoContentRepository.java
```diff
@@ -895,9 +892,13 @@ protected boolean createMissingFoldersCMIS(String fullPath) {
...
                     newFolderProps.put(PropertyIds.NAME, name);
-                    Folder newFolder = folder.createFolder(newFolderProps);
-                    Property property = newFolder.getProperty("alfcmis:nodeRef");
-                    newFolderRef = property.getValueAsString();
+                    try {
+                        Folder newFolder = folder.createFolder(newFolderProps);
+                        Property property = newFolder.getProperty("alfcmis:nodeRef");
+                        newFolderRef = property.getValueAsString();
+                    } catch (CmisContentAlreadyExistsException exc) {
+                        logger.info("Folder " + cleanPath + " already exists");
```
russdanner, line 900: This should not be an info.  Also, if you saw this in the logs you would have no idea what the context is, even as a warning it is not useful.  I'd remove it all together.

src/main/java/org/craftercms/studio/impl/v1/repository/alfresco/AlfrescoContentRepository.java
```diff
@@ -59,10 +59,7 @@
 import org.apache.chemistry.opencmis.commons.enums.BaseTypeId;
```
russdanner, line 59: I think this issue is also reproducible on disk repo.

PR #341, opened 2016-09-19
CRAFTERCMS-2138

dejan-brkic:
fixed issue


russdanner:
Need an explanation of what the diagnosis was and how this fixes the issue.

PR #390, opened 2016-11-09
CRAFTERCMS267

dejan-brkic:
Added cleaning path for db queries


russdanner (commented):
This code should really be out in a method

PR #395, opened 2016-11-16
CRAFTERCMS278

dejan-brkic:
fixed issue

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/deployment/PreviewDeployer.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/deployment/PreviewDeployer.java
@@ -207,12 +207,16 @@ public class PreviewDeployer implements Deployer {
     protected void syncFolder(String site, String path) {
         RepositoryItem[] children = contentRepository.getContentChildren(site, path, true);
 
-        for (RepositoryItem item : children) {
-            if (item.isFolder) {
-                syncFolder(site, item.path + "/" + item.name);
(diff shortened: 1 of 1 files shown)


src/main/java/org/craftercms/studio/impl/v1/deployment/PreviewDeployer.java
```diff
@@ -207,12 +207,16 @@ public void syncAllContentToPreview(RepositoryEventMessage message) throws Servi
...
 
-        for (RepositoryItem item : children) {
-            if (item.isFolder) {
-                syncFolder(site, item.path + "/" + item.name);
-            } else {
-                deployFile(site, contentService.getRelativeSitePath(site, item.path + "/" + item.name));
+        if (children != null) {
+            for (RepositoryItem item : children) {
+                if (item.isFolder) {
+                    syncFolder(site, item.path + "/" + item.name);
```
russdanner, line 213: We should create a Util called JoinPaths or something and get this  + "/" + junk out of our code.  and that util should correct for "//" or "///" or missing "/" scenarios

russdanner: I wont hold up this merge for that.... but we def should do that..

russdanner: I'll create a improvement ticket

PR #491, opened 2017-01-18
CRAFTERCMS issue #300

dejan-brkic:
CRAFTERCMS issue #300
Fixed datetime formatting issues


sumerjabri:
@dejan-brkic shouldn't this also go to version 3.0?

PR #588, opened 2017-03-17
CRAFTERCMS-561

dejan-brkic:
Fixed windows jgit path issue


src/main/java/org/craftercms/studio/impl/v1/service/security/DemoSecurityProvider.java
```diff
@@ -318,12 +332,24 @@ public boolean createGroup(String groupName, String description, String siteId)
...
+    @Override
     public List<Map<String, Object>> getUsersPerGroup(String site, String group, int start, int number) {
         // TODO: DB: Implement this ?
         return null;
     }
 
     @Override
+    public int getUsersPerGroupTotal(String site, String group) throws GroupNotFoundException {
+        // TODO: DB: Implement this ?
+        return 0;
```
sumerjabri: Why return 0?

src/main/java/org/craftercms/studio/impl/v1/service/security/TestSecurityProvider.java
```diff
@@ -266,12 +280,24 @@ public boolean createGroup(String groupName, String description, String siteId)
     }
 
     @Override
+    public int getGroupsPerSiteTotal(String site) {
+        // TODO: DB: Implement this ?
+        return 0;
```
sumerjabri: Why return 0?

src/main/java/org/craftercms/studio/impl/v1/service/security/TestSecurityProvider.java
```diff
@@ -266,12 +280,24 @@ public boolean createGroup(String groupName, String description, String siteId)
...
+    @Override
     public List<Map<String, Object>> getUsersPerGroup(String site, String group, int start, int number) {
         // TODO: DB: Implement this ?
         return null;
     }
 
     @Override
+    public int getUsersPerGroupTotal(String site, String group) throws GroupNotFoundException {
+        // TODO: DB: Implement this ?
+        return 0;
```
sumerjabri: Why return 0?

PR #612, opened 2017-03-30
CRAFTERCMS-592

dejan-brkic:
Ticket https://github.com/craftercms/craftercms/issues/592

Added validate session API
Fixed several bugs with session validation and session tokens


dejan-brkic:
Found blocking error - should not be merged

PR #681, opened 2017-05-08
Solve issue #794

cortiz:
Added a .keep file in the `/site/components` folder so it will be tracked and prevent users from creating components in the *empty* blueprint out of the box.


sumerjabri:
Please rename to. keep for consistency

PR #688, opened 2017-05-11
CRAFTERCMS-813

dejan-brkic:
Added monitoring API

Ticket:
https://github.com/craftercms/craftercms/issues/813


dejan-brkic:
CRAFTERCMS-802
Added code to avoid DB exception

Ticket:
https://github.com/craftercms/craftercms/issues/802

PR #731, opened 2017-06-06
CRAFTERCMS-925

dejan-brkic:
Fixed out of memory issue when uploading big files

Ticket:
https://github.com/craftercms/craftercms/issues/925


dejan-brkic:
CRAFTERCMS-927

Added mimetypes for .ftl (text/x-freemarker) and .groovy (text/x-groovy) files
Ticket:
https://github.com/craftercms/craftercms/issues/927

PR #747, opened 2017-06-16
CRAFTERCMS-913

dejan-brkic:
Fixed problems with configuration and token creation

Ticket:
https://github.com/craftercms/craftercms/issues/913


sumerjabri:
@dejan-brkic why is this still in the configuration: ` studio.security.file.configLocation: ${repository.diskImplementation.path}/security-config.xml`

PR #798, opened 2017-07-21
Add unit in description for Http session timeout

alhambrav:


sumerjabri:
@alhambrav value in minutes not seconds, please update.

PR #826, opened 2017-07-28
Updated default content monitoring configurations

jdrossl:
craftercms/craftercms#1128


sumerjabri:
@jdrossl I noticed the query allowing 2 days for emails to be sent, is that right? We want a single email reminder per content per period.

PR #892, opened 2017-10-11
CRAFTERCMS-1429

dejan-brkic:
Allow user to login with Studio credentials when LDAP authentication is setup but not reachable
Ticket:
https://github.com/craftercms/craftercms/issues/1429


src/main/java/org/craftercms/studio/impl/v1/service/security/DbWithLdapExtensionSecurityProvider.java
```diff
@@ -139,7 +140,7 @@ public User mapWithContext(DirContext dirContext, LdapEntryIdentification ldapEn
         User user;
         try {
             user = ldapTemplate.authenticate(ldapQuery, password, mapper);
-        } catch (EmptyResultDataAccessException e) {
+        } catch (EmptyResultDataAccessException | CommunicationException e) {
```
sumerjabri, line 143: Please separate into two different catches with two different logs that explain what happened and what you'll do. Right now communication error and user not found are the same in terms of how they logged.

PR #934, opened 2017-11-08
CRAFTERCMS-1575

dejan-brkic:
Replaced cherry picking with checking out content from specified commit id
Ticket:
https://github.com/craftercms/craftercms/issues/1575


dejan-brkic:
Cleaned up

PR #941, opened 2017-11-10
CRAFTERCMS-1583

dejan-brkic:
Randomize admin password generation

Ticket:
https://github.com/craftercms/craftercms/issues/1583


src/main/java/org/craftercms/studio/impl/v1/dal/DataSourceInitializerImpl.java
```diff
@@ -150,10 +153,20 @@ public void initDataSource() throws DatabaseUpgradeUnsupportedVersionException {
                         Reader reader = new InputStreamReader(is);
                         try {
                             sr.runScript(reader);
+
+                            if (isRandomAdminPasswordEnabled()) {
+                                String randomPassword = generateRandomPassword();
+                                String hashedPassword = CryptoUtils.hashPassword(randomPassword);
+                                String update = DB_QUERY_SET_ADMIN_PASSWORD.replace("{password}", hashedPassword);
+                                int result = statement.executeUpdate(update);
```
sumerjabri: Should we check the return value `result` and log/do something? If not, why have it?

dejan-brkic: We can remove it. It is driver dependent and usually is number of rows updated, but different vendors can omit the return value (result is 0).

sumerjabri: If it's of no use, omit please.

PR #968, opened 2017-11-30
Backwards compatibility for permissions configuration files

zac-poe:
Commit f933ff0da77edc2f7f718bbd3bbb3e5365164374 to master on Studio changed the Permissions Mapping configuration structure to remove the nested site element. This change provides backwards compatibility for existing sites using the previous configuration hierarchy, while still supporting the new, preferred hierarchy.


dejan-brkic (approved):
Backward compatibility

PR #985, opened 2017-12-14
CRAFTERCMS-1692

dejan-brkic:
Fixed bug where cut paste a scheduled page to be published does not clean workflow states

Ticket:
https://github.com/craftercms/craftercms/issues/1692


src/main/java/org/craftercms/studio/impl/v1/service/clipboard/ClipboardServiceImpl.java
```diff
@@ -117,6 +120,7 @@ protected void pasteItems(String site, String destinationPath, Set<ClipboardItem
                     // RDTMP_COPYPASTE
                     // CopyContent inteface is able to send status and new path yet
                     // newPath = contentService.moveContent(site, op.path, destinationPath);
```
sumerjabri: Please remove commented out code

sumerjabri (changes requested):

sumerjabri (approved):

PR #1001, opened 2018-01-04
CRAFTERCMS-1738

dejan-brkic:
Added configuration to ignore global gitignore

https://github.com/craftercms/craftercms/issues/1738


sumerjabri:
Per our conversation, we won't be doing this, instead, we will explain in docs what the best practices are.

PR #1009, opened 2018-01-17
Fix method call to doMonitoringForAllSites #1780

alhambrav:
Fix method call to doMonitoringForAllSites #1780


dejan-brkic (approved):
Not big changes, looks ok

PR #1031, opened 2018-02-05
CRAFTERCMS-1828

dejan-brkic:
Refactor locking

https://github.com/craftercms/craftercms/issues/1828

Don't approve this. It is just for Code Review


dejan-brkic:
Seems like it works fine. Tried with 5G file.

PR #1033, opened 2018-02-06
Solves 1815

cortiz:
Added configuration for ssh login/password and basic HTTP.


sumerjabri:
Per our conversation, please move to commons and rewire.

PR #1036, opened 2018-02-07
Moves reusable classes to Commons-git, Add commons git as dependency

cortiz:
Do not merge it before https://github.com/craftercms/commons/pull/120/  or build will fail.


dejan-brkic (approved):
In general looks good. Only one thing we need to keep watching for is clone result. It is possible that we don't get exception but clone failed. In that case failure reason is in clone result. Since we are not processing clone result and only error handling is with excwptions.

cortiz:
There is No Jgit check for what @dejan-brkic  describes.
 I did add some basic NPE prevention and make sure that the cloned folder exists, is a folder and user can RW on it.

PR #1037, opened 2018-02-07
CRAFTERCMS-1801

dejan-brkic:
Removed unused properties

https://github.com/craftercms/craftercms/issues/1801

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v1/util/StudioConfiguration.java
+++ b/src/main/java/org/craftercms/studio/api/v1/util/StudioConfiguration.java
@@ -32,10 +32,7 @@ public interface StudioConfiguration {
     String PUBLISHED_PATH = "studio.repo.sitePublishedPath";
     String BLUE_PRINTS_PATH = "studio.repo.blueprintsPath";
     String BOOTSTRAP_REPO = "studio.repo.bootstrapRepo";
-    String REPO_PREVIEW_ROOT_PATH = "studio.repo.previewRootPath";
     String REPO_REBUILD_METADATA_BATCH_SIZE = "studio.repo.rebuildMetadata.batchSize";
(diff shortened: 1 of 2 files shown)


src/main/java/org/craftercms/studio/api/v1/util/StudioConfiguration.java
```diff
@@ -115,9 +110,6 @@
 
     /** Security Service */
     String SECURITY_SESSION_TIMEOUT = "studio.security.sessionTimeout";
-    String SECURITY_URLS_TO_INCLUDE = "security.urlsToInclude";
```
sumerjabri, line 118: Why are we removing this? Right now things seem hardcoded here: https://github.com/craftercms/studio/blob/d534f989c957f8f21bc67f90ced47dcab78b2cc4/src/main/java/org/craftercms/studio/impl/v1/web/security/access/StudioGeneralAccessDecisionVoter.java#L37

We would like this to be configurable, no?

src/main/java/org/craftercms/studio/api/v1/util/StudioConfiguration.java
```diff
@@ -115,9 +110,6 @@
...
-    String SECURITY_URLS_TO_EXCLUDE = "studio.security.urlsToExclude";
```
sumerjabri, line 119: Same comment as above

src/main/java/org/craftercms/studio/api/v1/util/StudioConfiguration.java
```diff
@@ -115,9 +110,6 @@
...
-    String SECURITY_EXCEPTION_URLS = "studio.security.exceptionUrls";
```
sumerjabri, line 120: Same comment as above

sumerjabri (changes requested):

PR #1043, opened 2018-02-12
CRAFTERCMS-1833

dejan-brkic:
Fixed issue with delete dependency cascade

https://github.com/craftercms/craftercms/issues/1833

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/dependency/DependencyServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/dependency/DependencyServiceImpl.java
@@ -484,6 +484,17 @@ public class DependencyServiceImpl implements DependencyService {
 
     @Override
     public Set<String> getDeleteDepenencies(String site, String path) throws SiteNotFoundException, ContentNotFoundException, ServiceException {
+        // Check if site exists
+        if (!siteService.exists(site)) {
+            throw new SiteNotFoundException();
+        }
+
+        // Check if content exists
+        if (!contentService.contentExists(site, path)) {
+            throw new ContentNotFoundException();
+        }
+
+        logger.debug("Get delete dependencies for content - site " + site + " path " + path);
         List<String> paths = new ArrayList<String>();
         paths.add(path);
         Set<String> processedPaths = new HashSet<String>();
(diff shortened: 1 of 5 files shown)


src/main/java/org/craftercms/studio/impl/v1/service/dependency/DependencyServiceImpl.java
```diff
@@ -479,6 +482,82 @@ public void deleteSiteDependencies(@ValidateStringParam(name = "site") String si
...
+
+        // Step 2: collect all dependencies from DB
+        Set<String> dependencies = getItemDeleteDependencies(site, paths);
+        toRet.addAll(dependencies);
+        boolean doItAgain = false;
+
+        doItAgain = doItAgain || processedPaths.addAll(children);
+        doItAgain = doItAgain || processedPaths.addAll(dependencies);
+        if (doItAgain) {
+            List<String> pathsToProces = new ArrayList<String>();
```
sumerjabri: pathsToProcess not PathstoProces

src/main/java/org/craftercms/studio/impl/v1/service/dependency/DependencyServiceImpl.java
```diff
@@ -479,6 +482,82 @@ public void deleteSiteDependencies(@ValidateStringParam(name = "site") String si
...
+            pathsToProces.addAll(children);
+            pathsToProces.addAll(dependencies);
+            toRet.addAll(getDeleteDependenciesInternal(site, pathsToProces, processedPaths));
+        }
+
+        // Step 4: recursion ???
```
sumerjabri: Recursion is above in line 516, no? Why this comment?

src/main/java/org/craftercms/studio/impl/v1/service/dependency/DependencyServiceImpl.java
```diff
@@ -479,6 +482,82 @@ public void deleteSiteDependencies(@ValidateStringParam(name = "site") String si
...
+        return toRet;
+    }
+
+    private Set<String> getItemDeleteDependencies(String site, List<String> paths) {
```
sumerjabri: Method name doesn't convey what method actually does. Should this be called getItemDependencies or getFilteredItemDependencies?

Also, this is the third level in the call stack. Why this complexity. It's becoming hard to follow.

src/main/java/org/craftercms/studio/impl/v1/service/dependency/DependencyServiceImpl.java
```diff
@@ -479,6 +482,82 @@ public void deleteSiteDependencies(@ValidateStringParam(name = "site") String si
...
+            filtered.addAll(deps);
+            filtered = filterItemSpecificDependencies(filtered);
+            exitCondition = !toRet.addAll(filtered);
+            paths.clear();
+            paths.addAll(deps);
+        } while (!exitCondition);
+        return toRet;
+    }
+
+    private Set<String> getAllChildrenRecursively(String site, List<String> paths) {
```
sumerjabri: Doesn't this method exist somewhere else as some form of getTree? Or is this here because it returns a simple list of paths?

sumerjabri (changes requested):
Ping me to discuss.

dejan-brkic:
Added requested changes

PR #1073, opened 2018-03-16
CRAFTERCMS-1946

dejan-brkic:
Updated default regex and match index

https://github.com/craftercms/craftercms/issues/1946


sumerjabri (changes requested):
Don't you need () around the regex?

dejan-brkic:
There is no need for () in the regex. That is why index is set to 0 by default

PR #1079, opened 2018-03-26
CRAFTERCMS-1970

dejan-brkic:
Added sync blueprints feature

https://github.com/craftercms/craftercms/issues/1970

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
@@ -137,11 +137,13 @@ import static org.craftercms.studio.api.v1.constant.GitRepositories.SANDBOX;
 import static org.craftercms.studio.api.v1.constant.StudioConstants.BOOTSTRAP_REPO_GLOBAL_PATH;
 import static org.craftercms.studio.api.v1.constant.StudioConstants.BOOTSTRAP_REPO_PATH;
 import static org.craftercms.studio.api.v1.constant.StudioConstants.FILE_SEPARATOR;
+import static org.craftercms.studio.api.v1.util.StudioConfiguration.BLUE_PRINTS_PATH;
 import static org.craftercms.studio.api.v1.util.StudioConfiguration.BOOTSTRAP_REPO;
 import static org.craftercms.studio.api.v1.util.StudioConfiguration.REPO_PUBLISHED_COMMIT_MESSAGE;
 import static org.craftercms.studio.api.v1.util.StudioConfiguration.REPO_SANDBOX_BRANCH;
 import static org.craftercms.studio.api.v1.util.StudioConfiguration.SECURITY_CIPHER_KEY;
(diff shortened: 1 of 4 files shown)


src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
```diff
@@ -850,6 +852,31 @@ public void bootstrap() throws Exception {
...
+                } catch (GitAPIException err) {
+                    logger.error("error creating initial commit for global configuration", err);
+                }
+            } else {
+                // rsync blueprints
+                String bootstrapFolderPath = this.ctx.getRealPath(FILE_SEPARATOR + BOOTSTRAP_REPO_PATH + FILE_SEPARATOR + BOOTSTRAP_REPO_GLOBAL_PATH + FILE_SEPARATOR + studioConfiguration.getProperty(BLUE_PRINTS_PATH));
+                Path source = java.nio.file.FileSystems.getDefault().getPath(bootstrapFolderPath);
+                Path globalConfigPath = helper.buildRepoPath(GitRepositories.GLOBAL);
+                Path blueprintsPath = Paths.get(globalConfigPath.toAbsolutePath().toString(), studioConfiguration.getProperty(BLUE_PRINTS_PATH));
+                String[] cmd = new String[] {"rsync", "-avz", "--delete", source.toAbsolutePath().toString() + FILE_SEPARATOR, blueprintsPath.toAbsolutePath().toString() };
```
sumerjabri, line 865: Will this work on Windows?

sumerjabri: How about we delete and recreate the blueprints folder if something has changed?

src/main/webapp/repo-bootstrap/global/blueprints/website_editorial/test2.text
```diff
@@ -0,0 +1,3 @@
+test
```
sumerjabri: Do we want this file committed?

src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
sumerjabri: Perhaps a checksum file or a date file and we check/decide.

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
dejan-brkic: checksum file or timestamp file is a nice idea, the question is how do we automate it (if we rely on the user to update it, we might get into a problem if he forgets)

sumerjabri: the idea is to create a signature-file based on the war file's build number when starting up, we look at that file and if it's different from our build number, we delete the folder for the BPs and create them again and update the file

PR #1112, opened 2018-04-30
CRAFTERCMS-2072

dejan-brkic:
Added missing service call

https://github.com/craftercms/craftercms/issues/2072


src/main/webapp/default-site/scripts/rest/api/1/site/get-canned-message.get.groovy
```diff
@@ -0,0 +1,24 @@
+/*
```
sumerjabri, line 1: Fix copyright please

sumerjabri (changes requested):

PR #1126, opened 2018-05-08
CRAFTERCMS-2078

dejan-brkic:
Added caching for get user profile and get user groups calls

https://github.com/craftercms/craftercms/issues/2078

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
@@ -98,14 +98,6 @@ public class SecurityServiceImpl implements SecurityService {
     public Map<String,String> getUserProfile(final @ValidateStringParam(name = "user") String user) {
         StudioCacheContext cacheContext = new StudioCacheContext(CStudioConstants.CACHE_USERS_SCOPE, true);
         CacheService cacheService = cacheTemplate.getCacheService();
-        generalLockService.lock(cacheContext.getId());
-        try {
-            if (!cacheService.hasScope(cacheContext)) {
-                cacheService.addScope(cacheContext);
-            }
-        } finally {
-            generalLockService.unlock(cacheContext.getId());
-        }
 
         Map<String, String> userProfile = cacheTemplate.getObject(cacheContext, new Callback<Map<String, String>>() {
             @Override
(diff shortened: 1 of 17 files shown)


src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
```diff
@@ -566,6 +616,7 @@ public void reloadConfiguration(@ValidateStringParam(name = "site") String site)
             } else {
                 cacheService.addScope(cacheContext);
             }
+            cacheService.removeScope(usersCacheContext);
```
sumerjabri: Why are we removing the scope instead of clearing it?

src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
```diff
@@ -95,8 +95,25 @@ public String getCurrentToken() {
...
+        } finally {
+            generalLockService.unlock(cacheContext.getId());
+        }
+
+        Map<String, String> userProfile = cacheTemplate.getObject(cacheContext, new Callback<Map<String, String>>() {
+            @Override
+            public Map<String, String> execute() {
+                return securityProvider.getUserProfile(user);
+            }
+        }, "###USERS###", user, "profile");
```
sumerjabri: Please use a constant and not a literal string.

sumerjabri (changes requested):
When will this new cache scope be cleared and how? Ping me to discuss tomorrow.

src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
```diff
@@ -151,28 +168,47 @@ public PermissionsConfigTO execute() {
...
-    protected void addGlobalUserRoles(String user, Set<String> roles, PermissionsConfigTO rolesConfig) {
-        Set<String> groups = securityProvider.getUserGroups(user);
+    protected void addGlobalUserRoles( final String user, Set<String> roles, PermissionsConfigTO
+            rolesConfig) {
+
+        StudioCacheContext cacheContext = new StudioCacheContext(CStudioConstants.CACHE_USERS_SCOPE, true);
+        CacheService cacheService = cacheTemplate.getCacheService();
+        generalLockService.lock(cacheContext.getId());
+        try {
+            if (!cacheService.hasScope(cacheContext)) {
```
sumerjabri: Please put the creation of the scope in an init/constructor and the removal of the scope in a destructor or similar and just use the cache without checking/rechecking/re-creating every time.

sumerjabri (changes requested):

PR #1146, opened 2018-05-21
CRAFTERCMS-2165

dejan-brkic:
Updated param name to align with other APIs

https://github.com/craftercms/craftercms/issues/2165


sumerjabri (changes requested):
Shouldn't this be backward compatible too?

PR #1148, opened 2018-05-24
Feature/2177

alhambrav:
### Ticket reference or full description of what's in the PR ###
Add dependency resolver configuration in "Site Config" -> "Configuration" https://github.com/craftercms/craftercms/issues/2177


dejan-brkic (approved):
Everything looks ok. Regular expressions are the most sensitive and it is expected that we will change them slightly as our knowledge base grows

avasquez614 (approved):

PR #1154, opened 2018-06-06
Group name can now be extracted also with the LDAP site ID regex

avasquez614:
Group name can now be extracted also with the LDAP site ID regex, to allow for users to have specific site/groups.

Ticket craftercms/craftercms#2150


dejan-brkic (approved):
Everything seems to be in line with previous use cases

PR #1160, opened 2018-06-15
CRAFTERCMS-2207

dejan-brkic:
Added support for sandbox branch per site configuration

https://github.com/craftercms/craftercms/issues/2207

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
@@ -997,11 +997,7 @@ public class GitContentRepository implements ContentRepository, ServletContextAw
             // update site name variable inside config files
             toReturn = helper.updateSitenameConfigVar(site);
         }
-/*
-        if (toReturn) {
-            toReturn = helper.checkoutSandboxBranch(site);
-        }
-*/
+
         if (toReturn) {
             // commit everything so it is visible
             toReturn = helper.performInitialCommit(site, INITIAL_COMMIT);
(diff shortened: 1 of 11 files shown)


src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
```diff
@@ -995,7 +997,11 @@ public boolean createSiteFromBlueprint(String blueprintName, String site) {
             // update site name variable inside config files
             toReturn = helper.updateSitenameConfigVar(site);
         }
-
+/*
```
sumerjabri: Let's not leave commented out code.

src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
```diff
@@ -1071,7 +1080,7 @@ public void initialPublish(String site, String environment, String author, Strin
                 // checkout environment branch
                 logger.debug("Checkout environment branch " + environment + " for site " + site);
                 try {
-                    git.checkout().setCreateBranch(true).setForce(true).setStartPoint("master")
```
sumerjabri, line 1074: Just a comment: this was a bug, no? It didn't honor the config.

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
dejan-brkic: I don't think it was a bug. That is published repo. The process was to pull from sandbox into published master and from there to deploy content to the environment branch. With this feature, there is an additional obstacle I had to take care off. We can discuss it for further improvements.

russdanner:
@alhambrav to configure this:

In config/studio/site-config.xml add/change:
``` <sandbox-branch>master</sandbox-branch>```

PR #1177, opened 2018-07-06
CRAFTERCMS-2217

dejan-brkic:
Added submission comment as API parameter

https://github.com/craftercms/craftercms/issues/2217


russdanner:
confirmed

PR #1181, opened 2018-07-13
Working in Crafter Studio API 3:

avasquez614:
* Added possible response codes.

Ticket craftercms/craftercms#2275


src/main/api/3/studio-api-3.yaml
```diff
@@ -613,8 +617,34 @@ components:
         code:
           type: integer
           format: int32
+          description: |
+            A response code from one of the following:
```
sumerjabri: Can we find a way to say, it's value, description, remedial action, help URL (use the proper text headings in the sheet). Then for the values, we comma delimit?

sumerjabri (changes requested):

src/main/api/3/studio-api-3.yaml
avasquez614: I'll try to do a little more research. It's basically re-creating the table in the sheet right? If we can where should we put it?

sumerjabri (approved):

PR #1187, opened 2018-07-19
CRAFTERCMS-2283

dejan-brkic:
Prevent NPE

https://github.com/craftercms/craftercms/issues/2283


russdanner (commented):
Any idea why this would be null and if it should be ingored without a warning?

dejan-brkic:
Should not be null. The column is nullable in DB so there is a possibility that can receive null from DB

PR #1235, opened 2018-09-03
CRAFTERCMS-2412

dejan-brkic:
Fixed error when site id is null and total, offset, limit being 0 in response

https://github.com/craftercms/craftercms/issues/2412

### DIFF
--- a/src/main/java/org/craftercms/studio/controller/rest/v2/UsersController.java
+++ b/src/main/java/org/craftercms/studio/controller/rest/v2/UsersController.java
@@ -72,9 +72,9 @@ public class UsersController {
      */
     @GetMapping("/api/2/users")
     public StudioResponseBody getAllUsers(@RequestParam(value = "siteId", required = false) String siteId,
-                                          @RequestParam(value = "offset", required = false) int offset,
-                                          @RequestParam(value = "limit", required = false) int limit,
-                                          @RequestParam(value = "sort", required = false) String sort) {
+                      @RequestParam(value = "offset", required = false, defaultValue = "0") int offset,
+                      @RequestParam(value = "limit", required = false, defaultValue = "10") int limit,
+                      @RequestParam(value = "sort", required = false, defaultValue = StringUtils.EMPTY) String sort) {
         List<User> users = null;
         int total = 0;
         if (StringUtils.isEmpty(siteId)) {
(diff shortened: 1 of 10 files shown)


src/main/java/org/craftercms/studio/controller/rest/v2/UsersController.java
```diff
@@ -75,10 +75,22 @@ public StudioResponseBody getAllUsers(@RequestParam(value = "siteId", required =
                                           @RequestParam(value = "offset", required = false) int offset,
```
avasquez614: This should be an Integer since it's optional (can be null).

src/main/java/org/craftercms/studio/controller/rest/v2/UsersController.java
```diff
@@ -75,10 +75,22 @@ public StudioResponseBody getAllUsers(@RequestParam(value = "siteId", required =
...
                                           @RequestParam(value = "limit", required = false) int limit,
```
avasquez614: This should be an Integer since it's optional (can be null).

avasquez614 (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/security/UserServiceImpl.java
```diff
@@ -24,6 +24,7 @@
 import org.craftercms.studio.api.v1.log.Logger;
 import org.craftercms.studio.api.v1.log.LoggerFactory;
 import org.craftercms.studio.api.v2.dal.GroupDAO;
+import org.craftercms.studio.api.v2.dal.QueryParameterNames;
```
sumerjabri, line 27: Remove unused import.

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/controller/rest/v2/UsersController.java
avasquez614: @dejan-brkic I see that in the `GroupsController` you defined a default for offset instead. Shouldn't you do the same here?

src/main/java/org/craftercms/studio/controller/rest/v2/UsersController.java
avasquez614: @dejan-brkic I see that in the `GroupsController` you defined a default for limit instead. Shouldn't you do the same here?

avasquez614 (changes requested):

sumerjabri:
@dejan-brkic are you okay with the comments or do you want a meeting to resolve them?

dejan-brkic:
Sorry, didn't saw your request for change on unused import. No need for meeting

avasquez614 (approved):

sumerjabri (approved):

PR #1245, opened 2018-09-06
CRAFTERCMS-2408

dejan-brkic:
Removed preview sync on content service actions. Now it syncs only on repo sync (every 10 secs)

https://github.com/craftercms/craftercms/issues/2408


sumerjabri:
Won't accept per our conversation. We must have a more general approach that still gives the user an immediate index.

PR #1250, opened 2018-09-07
CRAFTERCMS-2386

dejan-brkic:
Fixed issue with role mappings not updated on upgrade. 
Fixed issue with group names not being lower case. 
Blueprints updated.
Adding default groups on upgrade.

https://github.com/craftercms/craftercms/issues/2386

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/dal/DataSourceInitializerImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/dal/DataSourceInitializerImpl.java
@@ -93,6 +93,8 @@ public class DataSourceInitializerImpl implements DataSourceInitializer {
     private final static String DB_VERSION_3_0_0 = "3.0.0";
     private final static String DB_VERSION_2_5_X = "2.5.x";
     private final static String DB_VERSION_3_0_X = "3.0";
+    private final static String DB_VERSION_3_1_0 = "3.1.0";
+    private final static String DB_VERSION_3_1_0_1 = "3.1.0.1";
 
     /**
      * Database queries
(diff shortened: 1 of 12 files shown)


src/main/resources/crafter/studio/database/upgrade-3.1.0.1.sql
```diff
@@ -0,0 +1,19 @@
+UPDATE `group`
+SET group_name = LCASE(group_name) ;
+
+INSERT IGNORE INTO `group` (record_last_updated, org_id, group_name, group_description)
```
avasquez614: Wouldn't this insert the groups again? Why is it not an UPDATE?

dejan-brkic: it will not insert them again. They were never inserted on the upgrade process

src/main/java/org/craftercms/studio/impl/v2/dal/DataSourceInitializerImpl.java
```diff
@@ -243,12 +248,14 @@ public void initDataSource()  throws DatabaseUpgradeUnsupportedVersionException,
                             logger.error("Integrity validator error after running upgrade DB scripts", e);
                         }
 
-                        if (dbVersion.startsWith(DB_VERSION_3_0_X)) {
+                        if (initialDbVersion.startsWith(DB_VERSION_3_0_X) ||
+                                initialDbVersion.equals(DB_VERSION_3_1_0) ||
+                                initialDbVersion.equals(DB_VERSION_3_1_0_1)) {
```
jdrossl, line 253: Why is `DB_VERSION_3_1_0_1` needed? are you planning on adding a new field for each update to the db?

dejan-brkic: no, just these two are affected besides 3.0.x

avasquez614: Every time the DB schema is updated, even if the version has not released, the DB version is updated with a snapshot version. The latest one is 3.1.0.2.

avasquez614 (approved):

PR #1256, opened 2018-09-10
CRAFTERCMS-2391

dejan-brkic:
removed magic numbers for oganization id

https://github.com/craftercms/craftercms/issues/2391


sumerjabri (dismissed):
Rename DEFAULT_ORGANIZATION_ID to ROOT_ORGANIZATION_ID.

Also, couldn't we have started at 0 instead of 1 for the root?

PR #1262, opened 2018-09-11
CRAFTERCMS-1727

dejan-brkic:
Added studio id field and internal service

https://github.com/craftercms/craftercms/issues/1727

### DIFF
--- a/src/main/webapp/default-site/scripts/rest/api/1/server/get-ui-resource-override.get.groovy
+++ b/src/main/webapp/default-site/scripts/rest/api/1/server/get-ui-resource-override.get.groovy
@@ -20,7 +20,5 @@ if(resource != null) {
 }
 else {
 	response.setStatus(404)
-    def instance = applicationContext.get("instanceService")
-
-	return instance.getInstanceId() //"NOT FOUND"
+	return "NOT FOUND"
 }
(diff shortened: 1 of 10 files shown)


sumerjabri (changes requested):
please fix the Codacy issues that can be fixed.

src/main/webapp/default-site/scripts/rest/api/1/server/get-ui-resource-override.get.groovy
```diff
@@ -20,5 +20,7 @@ if(resource != null) {
 }
 else {
 	response.setStatus(404)
-	return "NOT FOUND"
+    def instance = applicationContext.get("instanceService")
```
sumerjabri: Is this related to the instance ID?

sumerjabri (changes requested):

src/main/webapp/default-site/scripts/rest/api/1/server/get-ui-resource-override.get.groovy
sumerjabri: @dejan-brkic ^

dejan-brkic: I have removed that with the new commit. It was a mistake.

sumerjabri (approved):

PR #1275, opened 2018-09-14
Improvement for upgrade process:

jdrossl:
- Move users from and unused group and then delete it

https://github.com/craftercms/craftercms/issues/2473

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v2/upgrade/UpgradeConstants.java
+++ b/src/main/java/org/craftercms/studio/api/v2/upgrade/UpgradeConstants.java
@@ -24,7 +24,7 @@ package org.craftercms.studio.api.v2.upgrade;
  */
 public interface UpgradeConstants {
 
-    String CURRENT_DB_VERSION = "3.1.0.3";
+    String CURRENT_DB_VERSION = "3.1.0.4";
 
     String DB_VERSION_2_5_X = "2.5.x";
     String DB_VERSION_3_0_X = "3.0";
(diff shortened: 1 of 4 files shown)


avasquez614 (approved):

sumerjabri (changes requested):
Shouldn't this go into a new upgrade script 3.1.0.3 or at least into 3.1.0.2 after the `system_admin` group is officially created?
I know you're banking on the ID being 1, but isn't that also risky?
Why not do it after the group is created and using the group name to move users and delete the unused group. With name, we can't go wrong.

dejan-brkic (changes requested):
1. Every time new change in the database is added must go with the new update script and increased patch version. Only if changes did not get to the public repository (`develop`) you can merge them to the latest update script.

2. Delete cannot go by id because the group was already removed from create DB script. Delete statement must go by group name. One delete statement on `group` table should be enough because of foreign key cascade constraint.

3. The update statement is incorrect for the same reason as point 2. It is possible that the system does not have the group.

jdrossl:
@sumerjabri @dejan-brkic
1. does it make sense to add a new version just for content changes? we don't need to alter any table or column for this.

2 & 3. I'll change to avoid using ids, but we still need the update in case any user is still using that group.

dejan-brkic:
Anyone who has passed to a newer version than the one where you want to apply change will not pick up update.
The update is incorrect because it is blindly assigning all users from a group id=2 to group id=1 which is system admin group (full access)

sumerjabri:
@jdrossl I agree with Dejan. Please create a new SQL for this. Please use group names.

sumerjabri (approved):

sumerjabri:
@dejan-brkic final review please

dejan-brkic (approved):
Looks ok

PR #1284, opened 2018-09-17
Added /user/logout/url API

avasquez614:
Ticket craftercms/craftercms#2470


sumerjabri (changes requested):
Fix the Codacy issue

sumerjabri (changes requested):
Let's move the config to YAML and keep it as a single configuration item as we discussed.

sumerjabri (approved):

PR #1296, opened 2018-09-28
 Added role mapping for system_admin in all blueprints

avasquez614:
Ticket craftercms/craftercms#2514


avasquez614:
After discussion, we won't follow this approach.

PR #1297, opened 2018-10-02
CRAFTERCM-2383

dejan-brkic:
Applied new security pattern for API 2 services.
Introduced authentication chain

https://github.com/craftercms/craftercms/issues/2383


src/main/java/org/craftercms/studio/controller/rest/v2/GroupsController.java
```diff
@@ -72,18 +75,27 @@ public ResponseBody getAllGroups(
...
+            ResponseBody responseBody = new ResponseBody();
+            PaginatedResultList<GroupTO> result = new PaginatedResultList<>();
+            result.setTotal(total);
+            result.setOffset(offset);
+            result.setLimit(CollectionUtils.isEmpty(groups) ? 0 : groups.size());
+            result.setResponse(ApiResponse.OK);
+            result.setEntities(groups);
+            responseBody.setResult(result);
+            return responseBody;
+        } catch (OrganizationNotFoundException e) {
```
avasquez614: Can't we move this to ExceptionHandlers? That's where we're currently putting all our API exception handling code.

src/main/java/org/craftercms/studio/controller/rest/v2/GroupsController.java
```diff
@@ -112,14 +125,19 @@ public ResponseBody createGroup(@RequestBody Group group) throws GroupAlreadyExi
...
-        result.setEntity(updatedGroup);
-        responseBody.setResult(result);
+        try {
+            GroupTO updatedGroup = groupService.updateGroup(DEFAULT_ORGANIZATION_ID, group);
+
+            ResultOne<GroupTO> result = new ResultOne<>();
+            result.setResponse(ApiResponse.OK);
+            result.setEntity(updatedGroup);
+            responseBody.setResult(result);
+        } catch (GroupNotFoundException e) {
```
avasquez614: Can't we move this to ExceptionHandlers? That's where we're currently putting all our API exception handling code.

src/main/java/org/craftercms/studio/impl/v2/service/security/AuthenticationChainImpl.java
```diff
@@ -0,0 +1,121 @@
...
+    private ActivityService activityService;
+    private StudioConfiguration studioConfiguration;
+    private UserDAO userDao;
+    private GroupDAO groupDao;
+
+    public void init() {
+        Resource resource = new ClassPathResource(studioConfiguration.getProperty(
+                StudioConfiguration.CONFIGURATION_AUTHENTICATION_CHAIN_CONFIG_LOCATION));
+        try (InputStream in = resource.getInputStream()) {
+            Yaml yaml = new Yaml();
```
avasquez614, line 56: To specify the authentication chain, can't we just use the same studio config property `studio.security.type` and separate the chain by commas, eg. `studio.security.type: headers, ldap, db`?

src/main/webapp/repo-bootstrap/global/configuration/organization-permission-mappings-config.xml
```diff
@@ -0,0 +1,52 @@
...
+            <allowed-permissions>
+                <permission>Read</permission>
+                <permission>Write</permission>
+                <permission>Delete</permission>
+                <permission>Create Folder</permission>
+                <permission>Publish</permission>
+            </allowed-permissions>
+        </rule>
+    </role>
+    <role name="create-sites">
```
avasquez614: These are old permissions/roles that don't exist anymore. What is the point of this file?

src/main/webapp/repo-bootstrap/global/configuration/system-permission-mappings-config.xml
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
avasquez614: Sorry I don't understand the point of this file if we already have global permissions file where the permissions for the `system_admin` are defined.

avasquez614 (changes requested):
* Now that there's only UserTO/GroupTO, it's probably better to remove the TO suffix.
* I'm a little confused with the configuration changes. I thought we could use the existing properties, and change the security type to allow a comma-separated list so the chain can be defined, instead of using a new YAML file. I don't understand the new mapping files either. Maybe we can meet and discuss them?

src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
```diff
@@ -190,13 +195,13 @@ public Authentication getCurrentAuthentication() {
...
-        User u = securityProvider.getUserByIdOrUsername(-1, user);
+        UserTO u = securityProvider.getUserByIdOrUsername(-1, user);
         if (u != null) {
             toRet.put(KEY_USERNAME, user);
             toRet.put(KEY_FIRSTNAME, u.getFirstName());
             toRet.put(KEY_LASTNAME, u.getLastName());
             toRet.put(KEY_EMAIL, u.getEmail());
-            toRet.put(KEY_EXTERNALLY_MANAGED, u.isExternallyManaged());
+            toRet.put(KEY_EXTERNALLY_MANAGED, u.getExternallyManaged() > 0 ? true : false );
             String authenticationType = studioConfiguration.getProperty(SECURITY_TYPE);
```
avasquez614, line 205: Auth type is the type of the provider that authenticated the user. It shouldn't be the current security type in config, since even if it was something like `ldap`, a user can be authenticated by `db` in the end. Maybe we don't need to return this property here since it's only used by the new APIs

src/main/java/org/craftercms/studio/impl/v2/service/security/AuthenticationChainImpl.java
dejan-brkic: Security type will be removed in 3.1.1 (https://github.com/craftercms/craftercms/issues/2389)

src/main/webapp/repo-bootstrap/global/configuration/organization-permission-mappings-config.xml
dejan-brkic: I added that one long time ago, needs to be removed

src/main/webapp/repo-bootstrap/global/configuration/system-permission-mappings-config.xml
dejan-brkic: Same with this one, needs to be removed

src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
dejan-brkic: This is V1 service. Property is there as backward compatibility. Once we move all services to API 2 this will go away.

src/main/webapp/repo-bootstrap/global/configuration/system-permission-mappings-config.xml
avasquez614: K, will wait for the PR

src/main/webapp/repo-bootstrap/global/configuration/organization-permission-mappings-config.xml
avasquez614: K, will wait for the PR

avasquez614 (approved):

PR #1298, opened 2018-10-02
New Upgrade System

jdrossl:
https://github.com/craftercms/craftercms/issues/2355


src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/SiteVersionUpgradeOperation.java
```diff
@@ -0,0 +1,66 @@
...
+import org.craftercms.studio.api.v2.exception.UpgradeException;
+import org.springframework.core.io.ClassPathResource;
+import org.springframework.core.io.Resource;
+
+/**
+ * Implementation of {@link org.craftercms.studio.api.v2.upgrade.UpgradeOperation} that updates or adds the version
+ * file for the given site.
+ * @author joseross
+ */
+public class SiteVersionUpgradeOperation extends XsltFileUpgradeOperation {
```
avasquez614, line 34: Why extend `XsltFileUpgradeOperation`? It's not doing anything related to XSLT

avasquez614 (changes requested):

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/SiteVersionUpgradeOperation.java
jdrossl: There is an XSTL to update the version tag in any xml file, the difference is that this one adds the file if its missing from the repo.

avasquez614 (approved):

PR #1301, opened 2018-10-04
 Add urlrewrite.xml to all blueprints:

avasquez614:
* Added urlrewrite.xml to the config list of every blueprint.
* Added sample-urlrewrite.xml to every blueprint.


sumerjabri:
Shouldn't this also go into the upgrade mechanism :)

Let's discuss today.

sumerjabri:
Please use https://github.com/craftercms/craftercms/issues/2537 to add the necessary files to existing sites.

PR #1309, opened 2018-10-05
Added a description for the upgrade system in readme

jdrossl:
https://github.com/craftercms/craftercms/issues/2538

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,20 @@ In the current version of the upgrade system any change made to the blueprints i
 
 When there is a change in the database (structure or conent) there are two requirements:
 
-1. Update the create [script](https://github.com/craftercms/studio/blob/develop/src/main/resources/crafter/studio/database/createDDL.sql)
-2. Add an upgrade script in the [database folder](https://github.com/craftercms/studio/tree/develop/src/main/resources/crafter/studio/database)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,3 +4,104 @@ Crafter Studio
...
+
+In the current version of the upgrade system any change made to the blueprints in this [repository](https://github.com/craftercms/studio/tree/develop/src/main/webapp/repo-bootstrap/global/blueprints) will be automatically copied to the local global repository when Studio boots.
+
+*In the future this will change to follow the same approach used for the other components.*
+
+### Database & Configuration
+
+When there is a change in the database (structure or conent) there are two requirements:
+
+1. Update the create [script](https://github.com/craftercms/studio/blob/develop/src/main/resources/crafter/studio/database/createDDL.sql)
```
alhambrav: Would it be clearer if we say:
1. Update the create SQL script [script]

README.md
```diff
@@ -4,3 +4,104 @@ Crafter Studio
...
+2. Add an upgrade script in the [database folder](https://github.com/craftercms/studio/tree/develop/src/main/resources/crafter/studio/database)
```
alhambrav: Can we add a little more description here that the upgrade script is for the database?  It would also be nice if we have some sort of sample database upgrade script?

alhambrav (approved):

PR #1312, opened 2018-10-05
CRAFTERCMS-2541

dejan-brkic:
Changed order of operations when creating sites

https://github.com/craftercms/craftercms/issues/2541


jdrossl (commented):
@dejan-brkic I was wondering if this change also covers the case for create & push to a remote? is the push done before or after the upgrades?

jdrossl (approved):

PR #1313, opened 2018-10-05
Fixes for LDAP issues:

avasquez614:
* UserNotFoundException is not going to be thrown when updating a User with username and no ID.
* Fixed issues with LDAP provider that will throw exceptions when it should continue the chain.

Ticket craftercms/craftercms#2542


dejan-brkic (commented):
Can you verify these changes against this issue:
https://github.com/craftercms/craftercms/issues/2527

If these changes do not cause the error as described in the ticket, then it is approved.

avasquez614:
@dejan-brkic @sumerjabri double checked, I got the `Invalid username or password message` when logging with DB and LDAP

dejan-brkic (approved):
Ok

PR #1318, opened 2018-10-09
Added urlrewrite.xml to the upgrade pipeline:

avasquez614:
* sample-urlrewrite.xml will be put on existing sites.
* Added XSLT that will add the urlrewrite.xml to the config-list.xml.

### DIFF
--- a/src/main/resources/crafter/studio/upgrade/pipelines.yaml
+++ b/src/main/resources/crafter/studio/upgrade/pipelines.yaml
@@ -121,6 +121,9 @@ pipelines:
     - currentVersion: 3.0.0
       nextVersion: 3.1.0
       operations:
+        - type: addFileUpgrader
+          path: /config/studio/administration/samples/sample-urlrewrite.xml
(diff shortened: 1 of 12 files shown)


src/main/resources/crafter/studio/upgrade/pipelines.yaml
```diff
@@ -121,6 +121,9 @@ pipelines:
     - currentVersion: 3.0.0
```
sumerjabri, line 121: What if someone is flowing through 3.0.15 or 3.0.17

sumerjabri (changes requested):

src/main/resources/crafter/studio/upgrade/pipelines.yaml
avasquez614: Since sites < 3.1.0 don't have a `studio_version.xml`, the upgrader assumes a `3.0.0` version (just for the site).

sumerjabri (approved):

PR #1325, opened 2018-10-11
Added new AWS S3 API

jdrossl:
https://github.com/craftercms/craftercms/issues/2508


avasquez614 (changes requested):
When adding  any new API as API 2, please keep the interface of the old services, like S3Service, the same while adding a new service interface and model in package v2.

avasquez614:
@sumerjabri seems we're missing permissions for this new API. Should the user that's calling this API need `read/write content` permissions, or should we add new permissions called `read/write S3`?

PR #1326, opened 2018-10-11
CRAFTERCMS-2548

dejan-brkic:
Added logging and audit for authentication success and failure

https://github.com/craftercms/craftercms/issues/2548


src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
```diff
@@ -125,19 +125,40 @@
...
+            Map<String, String> extraInfo = new HashMap<String, String>();
+            extraInfo.put(DmConstants.KEY_CONTENT_TYPE, StudioConstants.CONTENT_TYPE_USER);
+            activityService.postActivity(getSystemSite(), username, ipAddress, activityType,
+                    ActivityService.ActivitySource.API, extraInfo);
+
+            logger.info("Failed to authenticate user " + username + " logging in from IP: " + ipAddress);
+
+            throw e;
+        }
+        if (StringUtils.isNotEmpty(toRet)) {
```
avasquez614, line 146: `toRet` seems always to be empty, after declaring it it's never assigned a value again

PR #1339, opened 2018-10-18
Fix to create all folders during webdav uploads

jdrossl:
https://github.com/craftercms/craftercms/issues/2581

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/webdav/WebDavServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/webdav/WebDavServiceImpl.java
@@ -102,7 +102,7 @@ public class WebDavServiceImpl implements WebDavService {
         throws
         WebDavException {
         WebDavProfile profile = profileReader.getProfile(site, profileId);
-        String listPath = profile.getBaseUrl();
+        String listPath = StringUtils.appendIfMissing(profile.getBaseUrl(),"/");
         MimeType filterType;
         Sardine sardine = SardineFactory.begin(profile.getUsername(), profile.getPassword());
         try {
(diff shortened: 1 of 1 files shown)


src/main/java/org/craftercms/studio/impl/v1/service/webdav/WebDavServiceImpl.java
```diff
@@ -180,22 +180,32 @@ public String upload(@ValidateStringParam(name = "site_id") final String site,
...
+
+                for(String folder : folders) {
+                    uploadUrl += "/" + folder;
+
+                    try {
+                        logger.debug("Trying to create folder {0}", uploadUrl);
+                        sardine.createDirectory(uploadUrl);
+                        logger.debug("Folder created");
+                    } catch (Exception e) {
+                        logger.debug("Folder already exists");
```
avasquez614: Wouldn't it be better to first check if the dir exists and then create it? Instead of trying to create it and then ignore the exception if it already exists. Also what if the exception is a connection issue?

avasquez614 (changes requested):

src/main/java/org/craftercms/studio/impl/v1/service/webdav/WebDavServiceImpl.java
jdrossl: This was implemented in this way because webdav only allows checking if files exist but not folders, so the only way was to try to create it and if an exception is thrown it means it already exists, about connectivity issues the current library doesn't provide specific exceptions to distinguish each case.

avasquez614: Understood. Maybe we can log the exception too and put a debug message of `Folder already exists or it couldn't be created`?

jdrossl: Actually I think there is a way to distinguish them, I didn't notice because its a runtime exception, I will improve this

avasquez614: Ok good

avasquez614: Also when  logging add the `uploadUrl` to the message. Eg. `"Folder " + uploadUrl + " created"`

avasquez614 (approved):

PR #1340, opened 2018-10-18
Fix to create all folders during webdav uploads

jdrossl:
https://github.com/craftercms/craftercms/issues/2581

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/webdav/WebDavServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/webdav/WebDavServiceImpl.java
@@ -102,7 +102,7 @@ public class WebDavServiceImpl implements WebDavService {
         throws
         WebDavException {
         WebDavProfile profile = profileReader.getProfile(site, profileId);
-        String listPath = profile.getBaseUrl();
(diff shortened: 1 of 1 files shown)


src/main/java/org/craftercms/studio/impl/v1/service/webdav/WebDavServiceImpl.java
```diff
@@ -180,22 +180,32 @@ public String upload(@ValidateStringParam(name = "site_id") final String site,
...
+
+                for(String folder : folders) {
+                    uploadUrl += "/" + folder;
+
+                    try {
+                        logger.debug("Trying to create folder {0}", uploadUrl);
+                        sardine.createDirectory(uploadUrl);
+                        logger.debug("Folder created");
+                    } catch (Exception e) {
+                        logger.debug("Folder already exists");
```
avasquez614: (Same as with https://github.com/craftercms/studio/pull/1339) Wouldn't it be better to first check if the dir exists and then create it? Instead of trying to create it and then ignore the exception if it already exists. Also what if the exception is a connection issue?

avasquez614 (changes requested):

avasquez614 (approved):

PR #1344, opened 2018-10-23
Clustering API

sumerjabri:
https://github.com/craftercms/craftercms/issues/2583

Added clustering API. Removed password field from AuthenticatedUser, and added notes for password fields so they're not returned to the caller.

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -1134,15 +1134,12 @@ paths:
               schema:
                 type: object
                 properties:
-                  result:
-                    type: object
-                    properties:
-                      response:
-                        $ref: '#/components/schemas/ApiResponse'
-                      entities:
-                        type: array
-                        items:
-                          $ref: '#/components/schemas/ClusterMember'
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+                  clusterMembers:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/ClusterMember'
         '400':
           $ref: '#/components/responses/BadRequest'
         '401':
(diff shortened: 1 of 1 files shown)


src/main/api/2/studio-api-2.yaml
```diff
@@ -1115,6 +1119,159 @@ paths:
...
+                      entities:
+                        type: array
+                        items:
+                          $ref: '#/components/schemas/ClusterMember'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
```
avasquez614, line 1148: I don't think we need to throw this error (404)?

src/main/api/2/studio-api-2.yaml
```diff
@@ -1115,6 +1119,159 @@ paths:
...
+      operationId: updateClusterMembers
+      requestBody:
+        description: Cluster members to update (upsert)
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                clusterMembers:
```
avasquez614: We're not wrapping currently the request objects. I would say remove `clusterMembers`

src/main/api/2/studio-api-2.yaml
```diff
@@ -1115,6 +1119,159 @@ paths:
...
+                        $ref: '#/components/schemas/ApiResponse'
+                      entities:
+                        type: array
+                        items:
+                          $ref: '#/components/schemas/ClusterMember'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
```
avasquez614: I don't think we need to throw this error (404)?

avasquez614 (dismissed):

src/main/api/2/studio-api-2.yaml
sumerjabri: What should we return if there is no cluster (single node with the DB table empty)?

avasquez614: You're right we should return 404

PR #1351, opened 2018-10-30
CRAFTERCMS-2593

dejan-brkic:
Implement cluster API

https://github.com/craftercms/craftercms/issues/2593

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v2/dal/ClusterDAO.java
+++ b/src/main/java/org/craftercms/studio/api/v2/dal/ClusterDAO.java
@@ -23,13 +23,46 @@ import java.util.Map;
 
 public interface ClusterDAO {
 
+    /**
+     * Get all cluster members from database
+     *
+     * @return List of cluster members
+     */
     List<ClusterMember> getAllMembers();
 
+    /**
+     * Update cluster member in the database
+     *
+     * @param member Cluster member to update
+     *
+     * @return number of affected rows
+     */
     int updateMember(ClusterMember member);
 
+    /**
+     * Add member to cluster in the database
+     *
+     * @param member Member to add
+     *
+     * @return number of affected rows
+     */
     int addMember(ClusterMember member);
 
+    /**
+     * Remove members from cluster in the database
+     *
+     * @param params Parameters for SQL query
+     *
+     * @return number of affected rows
+     */
     int removeMembers(Map params);
 
+    /**
+     * Get cluster member by id from database
+     *
+     * @param clusterMemberId Cluster member id
+     *
+     * @return Cluster member with given id
+     */
     ClusterMember getMemberById(long clusterMemberId);
 }
--- a/src/main/java/org/craftercms/studio/api/v2/dal/ClusterDAO.java
+++ b/src/main/java/org/craftercms/studio/api/v2/dal/ClusterDAO.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (C) 2007-2018 Crafter Software Corporation. All rights reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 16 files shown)


src/main/java/org/craftercms/studio/api/v2/service/cluster/internal/ClusterManagementServiceInternal.java
```diff
@@ -0,0 +1,36 @@
...
+
+package org.craftercms.studio.api.v2.service.cluster.internal;
+
+import org.craftercms.studio.api.v2.dal.ClusterMember;
+
+import java.util.List;
+
+public interface ClusterManagementServiceInternal {
+
+    List<ClusterMember> getAllMembers();
```
sumerjabri, line 32: Java doc this file, please.

src/main/java/org/craftercms/studio/impl/v2/service/cluster/internal/ClusterManagementServiceInternalImpl.java
```diff
@@ -0,0 +1,101 @@
...
+    @Override
+    public boolean addMember(ClusterMember clusterMember) {
+        if (StringUtils.isEmpty(clusterMember.getClusterId())) {
+            clusterMember.setClusterId(UUID.randomUUID().toString());
+        }
+        if (StringUtils.isEmpty(clusterMember.getClusterMemberName())) {
+            clusterMember.setClusterMemberName(UUID.randomUUID().toString());
+        }
+        if (StringUtils.isEmpty(clusterMember.getClusterMemberIp())) {
+            clusterMember.setClusterMemberIp("127.0.0.1");
```
sumerjabri: Why?
Shouldn't this cause an error?

src/main/java/org/craftercms/studio/impl/v2/service/cluster/internal/ClusterManagementServiceInternalImpl.java
```diff
@@ -0,0 +1,101 @@
...
+    @Override
+    public boolean updateMember(ClusterMember clusterMember) {
+        if (StringUtils.isEmpty(clusterMember.getClusterId())) {
+            clusterMember.setClusterId(UUID.randomUUID().toString());
+        }
+        if (StringUtils.isEmpty(clusterMember.getClusterMemberName())) {
+            clusterMember.setClusterMemberName(UUID.randomUUID().toString());
+        }
+        if (StringUtils.isEmpty(clusterMember.getClusterMemberIp())) {
+            clusterMember.setClusterMemberIp("127.0.0.1");
```
sumerjabri: Same issue as above.

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/cluster/internal/ClusterManagementServiceInternalImpl.java
dejan-brkic: Difference between DB model and API model. We need to discuss it. For now, that code is there not to break service

src/main/java/org/craftercms/studio/impl/v2/service/cluster/internal/ClusterManagementServiceInternalImpl.java
dejan-brkic: Difference between DB model and API model. We need to discuss it. For now, that code is there not to break service

dejan-brkic:
Aligned DB model with API model

src/main/java/org/craftercms/studio/api/v2/dal/ClusterDAO.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package org.craftercms.studio.api.v2.dal;
+
+import java.util.List;
+import java.util.Map;
+
+public interface ClusterDAO {
+
+    List<ClusterMember> getAllMembers();
```
sumerjabri, line 31: JavaDoc please

sumerjabri (changes requested):

sumerjabri (approved):

PR #1365, opened 2018-11-19
[studio] Add the new RTE to all bps #2631

jvega190:


alhambrav (changes requested):
In the sample rte configs, we have
```
<templates>
            <template>
                <title>Test template 1</title>
                <content>Test 1</content>
            </template>
            <template>
                <title>Test template 2</title>
                <content>Test 2</content>
            </template>
        </templates>
```
but, the above is not found in the actual configs... Should this be removed from the samples?

jvega190:
@alhambrav it is provided as an example of usage, but I don't think that is that common so I didn't add it to the configs. I kept it on the examples so it can be used if wanted. What do you think that is better?

alhambrav:
@jvega190 ah.... I guess we could keep it as long as it works when used in the config?  I haven't used it before, unfortunately.

alhambrav (approved):

sumerjabri:
We should consider:
- Video Center II <== should we do that one too
- Upgrade Manager <== this means it will add it to existing sites and BPs

PR #1375, opened 2018-11-28
CRAFTERCMS-2580

dejan-brkic:
Fixed issues found during QA environment setup

https://github.com/craftercms/craftercms/issues/2580


src/main/java/org/craftercms/studio/impl/v2/service/cluster/StudioClusterSyncJobImpl.java
```diff
@@ -71,7 +71,7 @@ public void run() {
                 params.put(CLUSTER_LOCAL_IP, localIp);
                 params.put(CLUSTER_STATE, ClusterMember.State.ACTIVE.toString());
                 List<ClusterMember> clusterMembers = clusterDAO.getOtherMembers(params);
-                logger.error("Number of active cluster members to sync with " + clusterMembers.size());
+                logger.debug("Number of active cluster members to sync with " + clusterMembers.size());
```
sumerjabri: Didn't we agree to:
- rename to become "Number of active cluster members: #"
- the number is the count of active members in the table?

sumerjabri (changes requested):

sumerjabri (approved):

PR #1381, opened 2018-12-03
Add CORS support for the API requests

jdrossl:
https://github.com/craftercms/craftercms/issues/2674

### DIFF
--- a/src/main/resources/crafter/studio/studio-config.yaml
+++ b/src/main/resources/crafter/studio/studio-config.yaml
@@ -524,9 +524,15 @@ studio.clustering.threadPool.maxPoolSize: 20
 ################################################################
 ##                             CORS                           ##
 ################################################################
+# Disable completely CORS headers
 studio.cors.disable: false
+# Value for the Access-Control-Allow-Origin header
 studio.cors.origins: '*'
+# Value for the Access-Control-Allow-Headers header
 studio.cors.headers: '*'
+# Value for the Access-Control-Allow-Methods header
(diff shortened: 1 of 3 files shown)


sumerjabri:
@jdrossl please also send a PR for `studio-config-override.yaml` in `craftercms` to allow the user to override/make changes.

jdrossl:
@sumerjabri I don't think we need to add them there to allow overriding, any property can be added if needed

src/main/resources/crafter/studio/studio-config.yaml
```diff
@@ -519,4 +519,14 @@ studio.clustering.threadPool.namePrefix: Studio_Clustering_Thread_Pool
...
 studio.clustering.threadPool.corePoolSize: 10
 # Thread pool max size
-studio.clustering.threadPool.maxPoolSize: 20
\ No newline at end of file
+studio.clustering.threadPool.maxPoolSize: 20
+
+################################################################
+##                             CORS                           ##
+################################################################
+studio.cors.disable: false
```
sumerjabri, line 531: Please document these fields (one comment per field above the field).

sumerjabri (changes requested):
Please also add these to the override file in `craftercms` with the comments that you're going to add here. Make sure it's commented out in the override.

The point of the override is to allow some orgs to make updates to tighten things up if required.

avasquez614 (approved):

sumerjabri (approved):

PR #1384, opened 2018-12-04
Add optional path for S3 file uploads

jdrossl:
https://github.com/craftercms/craftercms/issues/2663

Also fixed a version conflict with the aws dependencies from engine


avasquez614 (commented):
Is there any reason why we use path + filename instead of just the full path? I believe in other services like WebDav is the full path that's received.

jdrossl:
@avasquez614 webdav also uses a path, this is how all previous data sources work on Studio, they have an optional path for the uploads

avasquez614 (approved):

PR #1437, opened 2019-01-11
Feature to write audit logs to external file

apillai:
Feature to write audit logs to external file

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v1/util/StudioConfiguration.java
+++ b/src/main/java/org/craftercms/studio/api/v1/util/StudioConfiguration.java
@@ -225,6 +225,12 @@ public interface StudioConfiguration {
     String PUBLISHING_MANAGER_PUBLISHING_WITHOUT_DEPENDENCIES_ENABLED =
             "studio.publishingManager.publishingWithoutDependencies.enabled";
 
+    /** Audit Log */
+    String STUDIO_CONFIGURATION_AUDIT_LOG_PATH = "studio.configuration.audit.log.path";
+    String STUDIO_CONFIGURATION_AUDIT_LOG_FILE_NAME = "studio.configuration.audit.log.file.name";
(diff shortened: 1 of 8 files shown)


sumerjabri:
### Closing, reasoning below

- Discussed the PR and the intent was to push a copy of the audit table entries to ELK via `logstash` by pushing it to disk first.
  - The better approach is to point `logstash` to the audit table directly and have it push that to ELK
  - Example: https://discuss.elastic.co/t/how-to-connect-mysql-database-by-logstash/61004
```
input {
jdbc {
jdbc_driver_library => "C:/logstash/lib/mysql-connector-java-5.1.38-bin.jar"
jdbc_driver_class => "com.mysql.jdbc.Driver"
jdbc_connection_string => "jdbc:mysql://127.0.0.1:3306/lenderplatform_dev"
jdbc_user => "root"
jdbc_password => ""
statement => "SELECT * FROM lenderplatform_dev.loans"
jdbc_paging_enabled => "true"
jdbc_page_size => "50000"
}
}
output {
elasticsearch {
hosts => ["localhost:9200"]
index => "lenderplatform_dev"
document_type => "loans"
}
stdout { codec => json_lines }
}
```

- The current PR is pushing to `master` which is not permitted, code must flow into `develop` and a support branch if applicable
- Other issues:
  - We cannot switch off database logging
  - We cannot hardcode log4j as we're mostly switched to log4j2
  - We cannot have file logging defaulted to true
  - As things stand, log rotation will fail to rotate these new logs
  - Code reviews (Codacy) enumerates several issues that need to be addressed

If we *must* have Crafter log what's in the `audit` table to disk, the approach we would take is to have an agent pull that off the table and push it to a store (like S3). This can't be done by Studio as there are many Studios in a general cluster install. In this case, we're thinking `logstash` or similar can do a better job meeting the requirement of pushing audit logs to ELK.

PR #1440, opened 2019-01-17
CRAFTERCMS-2617

dejan-brkic:
update environment override update values for preview and authoring urls in runtime

https://github.com/craftercms/craftercms/issues/2617


sumerjabri (changes requested):
Please fix like we did for 3.0

dejan-brkic:
fixed

PR #1453, opened 2019-01-25
CRAFTERCMS-2804

dejan-brkic:
Get Available Blueprints API v2

https://github.com/craftercms/craftercms/issues/2804


sumerjabri (changes requested):
Please update to have the object `Blueprint` as a single object containing the whole data structure and we'll worry about OAS problems afterwards.

PR #1471, opened 2019-02-13
CRAFTERCMS-2817

dejan-brkic:
Fixed issue by escaping string in groovy pages

https://github.com/craftercms/craftercms/issues/2817


avasquez614:
@dejan-brkic I think we should use this class instead: https://commons.apache.org/proper/commons-text/javadocs/api-release/org/apache/commons/text/StringEscapeUtils.html. The other one has been deprecated.

PR #1486, opened 2019-02-26
CRAFTERCMS-2870

dejan-brkic:
Audit Log API v2 doc

https://github.com/craftercms/craftercms/issues/2870

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -1201,6 +1201,68 @@ paths:
                     $ref: '#/components/schemas/ApiResponse'
                   result:
                     $ref: '#/components/schemas/SearchResult'
+  /audit:
+    get:
+      tags:
+        - audit
+      summary: Get audit log
+      description: 'Required permission "AUDIT_LOG"'
+      operationId: getAuditLog
+      parameters:
+        - name: site_id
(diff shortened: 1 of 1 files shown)


src/main/api/2/studio-api-2.yaml
```diff
@@ -1201,6 +1201,68 @@ paths:
...
+      description: 'Required permission "AUDIT_LOG"'
+      operationId: getAuditLog
+      parameters:
+        - name: site_id
+          in: query
+          description: Site ID
+          required: true
+          schema:
+            type: string
+        - name: start
```
sumerjabri: The standard we're using is `offset` and `limit`

src/main/api/2/studio-api-2.yaml
```diff
@@ -1617,6 +1679,36 @@ components:
           type: object
           description: The values and counts of the facet
           additionalProperties: true
+    AuditLog:
+      type: object
+      properties:
+        modified_date:
+          type: string
+          description: Modified date of audit log entry
+        creation_date:
```
sumerjabri: `created` not `creation`

sumerjabri (changes requested):

sumerjabri (approved):

PR #1489, opened 2019-02-28
#2869

vestrada:
https://github.com/craftercms/craftercms/issues/2869 - [studio-ui] Create a global audit viewing tool #2869


sumerjabri:
We don't want to remove the audit capability from sites, we just want to add a global audit which lets you audit all, and limit the site audit to just events related to that site.

Ping me to discuss further.

vestrada:
@sumerjabri Fixed. I added again the audit config to the site conf page.

PR #1505, opened 2019-03-12
Remove 400 response from /audit as all parameters are optional

yacdaniel:
Remove the 400 Response from /Audit since all parameters are optional and a bad request cannot be received. 

https://github.com/craftercms/craftercms/issues/2870


dejan-brkic (commented):
There are integer parameters if I send string instead of an integer than it is a bad request, no?

yacdaniel:
@dejan-brkic You right, I'm able to get a `400` with `http://localhost:8080/studio/api/2/audit?offset=a`
Closing.

PR #1507, opened 2019-03-14
Publishing issue

dejan-brkic:
Fix issue publishing gets blocked when deleting dependency is followed by publishing content containing that dependency

Support issue


russdanner (approved):
This has been reviewed.

PR #1521, opened 2019-03-22
CRAFTERCMS-2982

dejan-brkic:
Update Audit API with filter options

https://github.com/craftercms/craftercms/issues/2982

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '2.0.21'
+  version: '2.0.22'
   title: Crafter Studio
   termsOfService: ''
   contact:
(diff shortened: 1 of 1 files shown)


src/main/api/2/studio-api-2.yaml
```diff
@@ -1215,7 +1215,7 @@ paths:
       parameters:
         - name: siteId
           in: query
-          description: Site ID
+          description: Site ID - optional for super admin user, otherwise required
```
sumerjabri, line 1218: Please add `site name`.
Please also add `cluster node id`.

sumerjabri (changes requested):
I also updated the design document and the filter-bys, let's discuss and agree on a final design. (The updates were necessary after a deeper discussion with Russ.)

dejan-brkic:
Updated API docs

sumerjabri (approved):

PR #1528, opened 2019-03-27
#2979

vestrada:
https://github.com/craftercms/craftercms/issues/2979 - Create site dialog ux enhancements #2979


sumerjabri (changes requested):
Why is bp_empty.png so high in resolution?

vestrada:
Done.

sumerjabri (approved):

PR #1549, opened 2019-04-04
Use the blueprint descriptor instead of the 'searchEngine' parameter

jdrossl:
https://github.com/craftercms/craftercms/issues/3033


sumerjabri (approved):

dejan-brkic (commented):
Why changing the order of operations when the site is created?

Why loading blueprint descriptor from site repository ?

jdrossl:
The order was changed to be able to load the blueprint descriptor from the repo, and it is done that way because without the parameter in the API it is the only way we can tell if a site needs to use Solr

PR #1551, opened 2019-04-05
CRAFTERCMS-2986

dejan-brkic:
Changed order of blueprint in create site dialog.

https://github.com/craftercms/craftercms/issues/2986


src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/BlueprintRenameFolderUpgradeOperation.java
```diff
@@ -0,0 +1,136 @@
...
+import static org.craftercms.studio.api.v1.util.StudioConfiguration.BLUE_PRINTS_PATH;
+import static org.craftercms.studio.api.v1.util.StudioConfiguration.REPO_BLUEPRINTS_UPDATED_COMMIT_MESSAGE;
+import static org.craftercms.studio.impl.v1.repository.git.GitContentRepositoryConstants.GIT_COMMIT_ALL_ITEMS;
+
+/**
+ * Implementation of {@link org.craftercms.studio.api.v2.upgrade.UpgradeOperation} that renames the blueprint in the
+ * global repository.
+ * @author Dejan Brkic
+ */
+public class BlueprintRenameFolderUpgradeOperation extends AbstractUpgradeOperation implements ServletContextAware {
```
jdrossl: - `AbstractUpgradeOperation` already implements `ServletContextAware`
- Wouldn't it be better if we make a generic `RenameFolderUpgradeOperation` in case we need it later? or even for files too

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/BlueprintRenameFolderUpgradeOperation.java
```diff
@@ -0,0 +1,136 @@
...
+    @Override
+    public void doInit(final Configuration config) {
+        blueprintFolderName = config.getString(CONFIG_KEY_BLUEPRINT_FOLDER_NAME);
+        newFolderName = config.getString(CONFIG_KEY_NEW_FOLDER_NAME);
+    }
+
+    @Override
+    public void execute(final String site) throws UpgradeException {
+        try {
+            GitContentRepositoryHelper helper =
```
jdrossl: Maybe we should try to use the inherited `getRepositoryPath(String site)` to:
- Avoid the security & user service dependencies? I remember it caused some issues because when the upgrade runs there is no logged user
- Make sure we can use it for both global & site repos too

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/BlueprintRenameFolderUpgradeOperation.java
```diff
@@ -0,0 +1,136 @@
...
+            if (blueprintFolder.exists()) {
+                Path newBlueprintPath = Paths.get(globalConfigPath.toAbsolutePath().toString(),
+                        studioConfiguration.getProperty(BLUE_PRINTS_PATH), newFolderName);
+                File newBlueprintFolder = newBlueprintPath.toFile();
+                FileUtils.deleteDirectory(newBlueprintFolder);
+                blueprintFolder.renameTo(newBlueprintFolder);
+            }
+
+            Repository globalRepo = helper.getRepository(site, GitRepositories.GLOBAL);
+            try (Git git = new Git(globalRepo)) {
```
jdrossl: We should use the inherited `commitAllChanges(String site)` for consistency

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/BlueprintRenameFolderUpgradeOperation.java
```diff
@@ -0,0 +1,136 @@
...
+
+    public static final String CONFIG_KEY_BLUEPRINT_FOLDER_NAME = "blueprintFolderName";
+    public static final String CONFIG_KEY_NEW_FOLDER_NAME = "newFolderName";
+
+    protected ServletContext servletContext;
+    protected ServicesConfig servicesConfig;
+    protected SecurityService securityService;
+    protected UserServiceInternal userServiceInternal;
+
+    protected String blueprintFolderName;
```
jdrossl: If we use full paths for configuration then it can be completely generic and not specific to global repo or blueprints

src/main/resources/crafter/studio/upgrade/pipelines.yaml
```diff
@@ -246,6 +246,33 @@ pipelines:
...
+    - currentVersion: 3.1.0.20
+      nextVersion: 3.1.0.21
+      operations:
+        - type: dbScriptUpgrader
+          filename: upgrade-3.1.0.20-to-3.1.0.21.sql
+          updateIntegrity: true
+        - type: blueprintRenameUpgrader
+          blueprintFolderName: website_editorial
+          newFolderName: 1000_website_editorial
+          commitDetails: Rename Editorial blueprint
```
jdrossl, line 258: The `commitDetails` property will be ignored if we don't use the inherited method for doing the commit

jdrossl (changes requested):

jdrossl:
@sumerjabri @dejan-brkic After doing the review I realized we might not even need a new operation, right now we always replace the blueprints folder in each upgrade so would it take care of the rename?

dejan-brkic:
Before adding new operation it was adding renamed blueprints without removing old blueprints

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/BlueprintRenameFolderUpgradeOperation.java
dejan-brkic: I took `BlueprintsUpgradeOperation` as the starting point. Has the same extends and implements. I will fix them both.
At the moment I don't see a use case that needs the generic operation.

jdrossl: Yes `BlueprintsUpgradeOperation` was left behind after some refactors, if you can update it also it would be great.

But I still think it would be better to make it as generic as possible so we don't end with a lot of specific classes and even if we don't needed it now we might need to rename folders or files inside the site repositories in the future, @sumerjabri what do you think about this?

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/BlueprintRenameFolderUpgradeOperation.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+public class BlueprintRenameFolderUpgradeOperation extends AbstractUpgradeOperation {
+
+    private static final Logger logger = LoggerFactory.getLogger(BlueprintRenameFolderUpgradeOperation.class);
+
+
+    public static final String CONFIG_KEY_BLUEPRINT_FOLDER_NAME = "blueprintFolderName";
+    public static final String CONFIG_KEY_NEW_FOLDER_NAME = "newFolderName";
+
+    protected ServletContext servletContext;
```
jdrossl: We don't need `servletContext` & `servicesConfig` any more

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/BlueprintRenameFolderUpgradeOperation.java
```diff
@@ -0,0 +1,94 @@
...
+    protected ServicesConfig servicesConfig;
+
+    protected String blueprintFolderName;
+    protected String newFolderName;
+
+    @Override
+    public void setServletContext(final ServletContext servletContext) {
```
jdrossl: We don't need `setServletContext` & `setServicesConfig` any more

src/main/resources/crafter/studio/studio-upgrade-context.xml
```diff
@@ -136,6 +136,13 @@
         <property name="userServiceInternal" ref="userServiceInternal" />
     </bean>
 
+    <bean id="blueprintRenameUpgrader" scope="prototype" parent="upgradeOperation"
+          class="org.craftercms.studio.impl.v2.upgrade.operations.BlueprintRenameFolderUpgradeOperation">
+        <property name="servicesConfig" ref="cstudioServicesConfig"/>
```
jdrossl: We don't need the properties any more

jdrossl (changes requested):

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/BlueprintRenameFolderUpgradeOperation.java
sumerjabri: Let's do a generic `RenameOperation` so we can rename anything. Any issues with that?

sumerjabri (changes requested):
See my inline comment.

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/BlueprintRenameFolderUpgradeOperation.java
```diff
@@ -0,0 +1,72 @@
...
+    @Override
+    public void doInit(final Configuration config) {
+        blueprintFolderName = config.getString(CONFIG_KEY_BLUEPRINT_FOLDER_NAME);
+        newFolderName = config.getString(CONFIG_KEY_NEW_FOLDER_NAME);
+    }
+
+    @Override
+    public void execute(final String site) throws UpgradeException {
+        try {
+            Path globalConfigPath = getRepositoryPath(StringUtils.EMPTY);
```
jdrossl: If we use the `site` parameter it can work for any repo not just the global repo

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/RenameUpgradeOperation.java
```diff
@@ -24,48 +24,54 @@
...
+                FileUtils.forceDelete(newF);
+            }
+            if (oldF.isDirectory()) {
+                FileUtils.moveDirectory(oldF, newF);
+            } else if (oldF.isFile()) {
+                FileUtils.moveFile(oldF, newF);
             }
 
-            commitAllChanges(StringUtils.EMPTY);
+            commitAllChanges(siteId);
```
jdrossl: Please use the `site` parameter from the method, the upgrade manager will send the right value

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/RenameUpgradeOperation.java
```diff
@@ -24,48 +24,54 @@
...
 
-    public static final String CONFIG_KEY_BLUEPRINT_FOLDER_NAME = "blueprintFolderName";
-    public static final String CONFIG_KEY_NEW_FOLDER_NAME = "newFolderName";
+    public static final String CONFIG_KEY_SITE = "site";
+    public static final String CONFIG_KEY_OLD_PATH = "oldPath";
+    public static final String CONFIG_KEY_NEW_PATH = "newPath";
 
-    protected String blueprintFolderName;
-    protected String newFolderName;
+    protected String siteId;
```
jdrossl: This property is not needed, se the note on the method

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/RenameUpgradeOperation.java
```diff
@@ -24,48 +24,54 @@
...
-                FileUtils.deleteDirectory(newBlueprintFolder);
-                blueprintFolder.renameTo(newBlueprintFolder);
+
+            Path oldP = Paths.get(getRepositoryPath(siteId).getParent().toAbsolutePath().toString(), oldPath);
+            Path newP = Paths.get(getRepositoryPath(siteId).getParent().toAbsolutePath().toString(), newPath);
+
+            File oldF = oldP.toFile();
+            File newF = newP.toFile();
+            if (newF.exists()) {
+                FileUtils.forceDelete(newF);
```
jdrossl: If the new path already exists it means the rename was done either by someone manually or by other upgrade, I think instead of deleting it (very dangerous) the class should end without doing any changes

jdrossl (changes requested):

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/RenameUpgradeOperation.java
dejan-brkic: I don't want it to be run for all sites

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/RenameUpgradeOperation.java
dejan-brkic: I don't want it to be run for all sites

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/RenameUpgradeOperation.java
dejan-brkic: Currently, without deleting, it will never execute because `BlueprintsUpgradeOperation` runs first and copies renamed blueprints.

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/RenameUpgradeOperation.java
jdrossl: it will not run for all sites, the method parameter indicates the current site being upgraded (or will be empty for the global)

dejan-brkic: A current site being upgraded means on startup will run for all sites

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/RenameUpgradeOperation.java
jdrossl: ok then it would be better to add a property `overwrite` that defaults to false, and we set it to true for this one in the yaml

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/RenameUpgradeOperation.java
jdrossl: the upgrade manager has a different pipeline for the existing sites, this one is in the `system` pipeline so it will only run for the global repo

jdrossl:
@dejan-brkic looks good, let's just chat about using the `site` parameter if you still have questions

jdrossl (approved):

sumerjabri (approved):

PR #1557, opened 2019-04-09
CRAFTERCMS-2986

dejan-brkic:
fix startup issue

https://github.com/craftercms/craftercms/issues/2986


jdrossl (changes requested):
Please also add a default value to make the property optional `overwrite = config.getBoolean(CONFIG_KEY_OVERWRITE, false);`

jdrossl (approved):

PR #1569, opened 2019-04-15
CRAFTERCMS-3058;CRAFTERCMS-3039

dejan-brkic:
Added new environment properties
https://github.com/craftercms/craftercms/issues/3058

Moved sample configuration files to the global repository
https://github.com/craftercms/craftercms/issues/3039


src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/DeleteUpgradeOperation.java
```diff
@@ -0,0 +1,72 @@
...
+import java.io.File;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+/**
+ * Implementation of {@link org.craftercms.studio.api.v2.upgrade.UpgradeOperation} that renames the blueprint in the
+ * global repository.
+ * @author Dejan Brkic
+ */
+public class DeleteUpgradeOperation extends AbstractUpgradeOperation {
```
avasquez614, line 35: Why does the comment say that it's to rename blueprints but the class name is delete? Should the comment be updated?

avasquez614 (changes requested):

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/DeleteUpgradeOperation.java
sumerjabri: I think it's copy/paste. @dejan-brkic please fix.

avasquez614 (approved):

PR #1576, opened 2019-04-23
CRAFTERCMS-3095

dejan-brkic:
Added default dependency resolver config

https://github.com/craftercms/craftercms/issues/3095

### DIFF
--- a/src/main/resources/crafter/studio/studio-config.yaml
+++ b/src/main/resources/crafter/studio/studio-config.yaml
@@ -108,6 +108,10 @@ studio.configuration.site.dataSources.configBasePath: /config/studio/data-source
 studio.configuration.site.dataSources.configFileName: cmis-config.xml
 # Destroy site context url for preview engine
 studio.configuration.site.preview.destroy.context.url: http://localhost:8080/api/1/site/context/destroy.json?crafterSite={siteName}
+# Location pattern where default dependency resolver specific configuration for is stored
+studio.configuration.default.dependencyResolver.configBasePath: /configuration/dependency
+# File name where default dependency specific configuration is stored
+studio.configuration.default.dependencyResolver.configFileName: resolver-config.xml
 # Location pattern where dependency resolver specific configuration for a site is stored
 studio.configuration.site.dependencyResolver.configBasePath: /config/studio/dependency
 # File name where dependency specific configuration is stored
--- a/src/main/resources/crafter/studio/upgrade/3.1.0.25/resolver-config.xml
+++ b/src/main/resources/crafter/studio/upgrade/3.1.0.25/resolver-config.xml
@@ -246,15 +246,15 @@
             <!-- how to identify items of this type -->
             <includes>
                 <!-- path pattern regexes (multiple) -->
-				<!-- include only textual files that can be scanned -->
-				<path-pattern>/static-assets/([^&lt;"'\)]+)\.css</path-pattern>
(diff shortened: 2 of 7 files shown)


src/main/resources/crafter/studio/upgrade/pipelines.yaml
```diff
@@ -326,6 +326,15 @@ pipelines:
               dest: configuration/samples/sample-urlrewrite.xml
             - src: crafter/studio/upgrade/3.1.0.24/configuration/samples/sample-webdav.xml
               dest: configuration/samples/sample-webdav.xml
+    - currentVersion: 3.1.0.24
+      nextVersion: 3.1.0.25
+      operations:
+        - type: dbScriptUpgrader
```
jdrossl: Maybe it would be better to update the version as the last operation, that way if something else fails the version will not be changed.

jdrossl (changes requested):

src/main/resources/crafter/studio/studio-config.yaml
```diff
@@ -108,6 +108,10 @@ studio.configuration.site.dataSources.configBasePath: /config/studio/data-source
 studio.configuration.site.dataSources.configFileName: cmis-config.xml
 # Destroy site context url for preview engine
 studio.configuration.site.preview.destroy.context.url: http://localhost:8080/api/1/site/context/destroy.json?crafterSite={siteName}
+# Location pattern where default dependency resolver specific configuration for is stored
+studio.configuration.default.dependencyResolver.configBasePath: /configuration/dependency
```
avasquez614, line 112: Shouldn't it be better to specify these properties as one, since we're only using the path?

src/main/resources/crafter/studio/upgrade/3.1.0.25/resolver-config.xml
```diff
@@ -0,0 +1,329 @@
...
+            </dependency-types>
+        </item-type>
+        <item-type>
+            <!-- name of type -->
+            <name>asset</name>
+            <!-- how to identify items of this type -->
+            <includes>
+                <!-- path pattern regexes (multiple) -->
+				<!-- include only textual files that can be scanned -->
+				<path-pattern>/static-assets/([^&lt;"'\)]+)\.css</path-pattern>
```
avasquez614: Too much indentation

src/main/webapp/repo-bootstrap/global/configuration/dependency/resolver-config.xml
```diff
@@ -0,0 +1,329 @@
...
+            </dependency-types>
+        </item-type>
+        <item-type>
+            <!-- name of type -->
+            <name>asset</name>
+            <!-- how to identify items of this type -->
+            <includes>
+                <!-- path pattern regexes (multiple) -->
+				<!-- include only textual files that can be scanned -->
+				<path-pattern>/static-assets/([^&lt;"'\)]+)\.css</path-pattern>
```
avasquez614: Too much indentation

avasquez614 (changes requested):

src/main/resources/crafter/studio/studio-config.yaml
dejan-brkic: All configuration files have properties like that (filename and path in separate properties)

russdanner (approved):
My review of this code is that:
- The global uses the new rule set (which I am good with)
- The code has been updated so that if a site is created without its own rule set, it goes to the global (which I am good with)

avasquez614 (approved):

PR #1601, opened 2019-05-27
Update support for AWS MediaConvert:

jdrossl:
- Migrate service to API2
- Improve service to return a list of urls for all configured outputs
- Use remote assets urls

https://github.com/craftercms/craftercms/issues/2959


src/main/api/2/studio-api-2.yaml
```diff
@@ -1088,6 +1088,55 @@ paths:
...
+                file:
+                  type: string
+                  format: binary
+                  description: The content of the file to upload
+              required:
+                - siteId
+                - inputProfileId
+                - outputProfileId
+                - file
+      responses:
```
sumerjabri, line 1124: No 404? What if the job profiles are not present?

sumerjabri (changes requested):

src/main/api/2/studio-api-2.yaml
```diff
@@ -1088,6 +1088,55 @@ paths:
...
+                properties:
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+                  item:
+                    $ref: '#/components/schemas/MediaConvertResult'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '500':
+          $ref: '#/components/responses/InternalServerError'
+
```
avasquez614, line 1142: Could there be a bad request response if the request body has required or missing fields?

avasquez614 (changes requested):

src/main/api/2/studio-api-2.yaml
jdrossl: Any error from AWS related services will be mapped to a 500 response

src/main/api/2/studio-api-2.yaml
jdrossl: The service will return a 400 response, I'll add it to the doc

avasquez614 (approved):

sumerjabri (approved):

PR #1607, opened 2019-05-30
Prevent malformed XML from stop the site upgrade

jdrossl:
https://github.com/craftercms/craftercms/issues/3177

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/AbstractContentTypeUpgradeOperation.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/AbstractContentTypeUpgradeOperation.java
@@ -148,6 +148,9 @@ public abstract class AbstractContentTypeUpgradeOperation extends AbstractConten
             } else {
                 return true;
             }
+        } catch (ExecutionException e) {
+            logger.error("Invalid XML file found at " + file, e);
+            return false;
         } catch (Exception e) {
             throw new UpgradeException("Error parsing xml file " + file, e);
         }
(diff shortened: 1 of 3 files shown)


src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/AbstractContentTypeUpgradeOperation.java
```diff
@@ -148,6 +148,9 @@ protected boolean shouldBeUpdated(final String site, final Path file) throws Upg
             } else {
                 return true;
             }
+        } catch (ExecutionException e) {
```
avasquez614, line 151: Shouldn't we just continue with any exception? Why are we stopping the whole pipeline if we can't parse an XML?

avasquez614 (changes requested):

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/AbstractContentTypeUpgradeOperation.java
jdrossl: I don't think it would be ok to skip an operation if it fails, the site would be left in an unknown state but it would look like it was updated, something that fails could cause even more errors in future upgrades or in the site itself

sumerjabri: We should log errors and continue

jdrossl: @sumerjabri should this be only for sites or for the db & config too?

sumerjabri: Log errors and continue, except for DB errors or global/system-wide errors.

avasquez614 (approved):

PR #1611, opened 2019-06-03
[studio-ui] Add a Date filter in the Studio search filters #3138

jvega190:
https://github.com/craftercms/craftercms/issues/3138

### DIFF
--- a/src/main/resources/crafter/studio/studio-config.yaml
+++ b/src/main/resources/crafter/studio/studio-config.yaml
@@ -647,18 +647,20 @@ studio.search.facets:
     field: lastEditedOn
     date: true
     ranges:
-      - label: older
-        to: now/d-365d
-      - label: within-a-year
-        from: now/d-365d
-      - label: within-6-months
-        from: now/d-180d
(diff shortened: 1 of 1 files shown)


src/main/resources/crafter/studio/studio-config.yaml
```diff
@@ -647,18 +647,20 @@ studio.search.facets:
...
-      - label: older
-        to: now/d-365d
-      - label: within-a-year
-        from: now/d-365d
-      - label: within-6-months
-        from: now/d-180d
-      - label: within-a-month
-        from: now/d-30d
-      - label: within-a-week
+      - label: in-last-day
```
rart: Change to say today. Need to update the other related PRs.

rart (approved):

src/main/resources/crafter/studio/studio-config.yaml
jvega190: fortunately there's a "today" translation in search bundle, so this is the only file that needs to change

rart (approved):

PR #1612, opened 2019-06-05
Make indexing for authoring always async

jdrossl:
https://github.com/craftercms/craftercms/issues/3152

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/deployment/PreviewDeployerImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/deployment/PreviewDeployerImpl.java
@@ -85,7 +85,7 @@ public class PreviewDeployerImpl implements PreviewDeployer {
 
     @EventListener(EVENT_PREVIEW_SYNC)
     public void onPreviewSync(PreviewEventContext context) {
-        doDeployment(context.getSite(), ENV_AUTHORING, context.isWaitTillDeploymentIsDone());
+        doDeployment(context.getSite(), ENV_AUTHORING, false);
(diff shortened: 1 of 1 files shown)


src/main/java/org/craftercms/studio/impl/v1/deployment/PreviewDeployerImpl.java
```diff
@@ -85,7 +85,7 @@ public void subscribeToPreviewSyncEvents() {
 
     @EventListener(EVENT_PREVIEW_SYNC)
     public void onPreviewSync(PreviewEventContext context) {
-        doDeployment(context.getSite(), ENV_AUTHORING, context.isWaitTillDeploymentIsDone());
```
avasquez614, line 88: Shouldn't we remove this property from the context too?

avasquez614 (changes requested):

src/main/java/org/craftercms/studio/impl/v1/deployment/PreviewDeployerImpl.java
jdrossl: Sometimes it is set to `false`, not sure what the explanation is for each use case: db sync & cluster sync

avasquez614: Ok, I thought the flag wasn't used somewhere else, my bad.

sumerjabri: I say we keep the option to wait.

avasquez614 (approved):

PR #1614, opened 2019-06-06
CRAFTERCMS-2587

dejan-brkic:
Active environment feature

https://github.com/craftercms/craftercms/issues/2587

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -1230,7 +1230,7 @@ paths:
                   blueprints:
                     type: array
                     items:
-                      $ref: '#/components/schemas/Blueprint'
+                      $ref: '#/components/schemas/Plugin'
         '500':
           $ref: '#/components/responses/InternalServerError'
 
@@ -1317,19 +1317,19 @@ paths:
               type: string
         - name: includeParameters
           in: query
-          description: Include parameters into result set if `true`. Default `false` 
+          description: Include parameters into result set if `true`. Default `false`
           required: false
           schema:
             type: boolean
         - name: dateFrom
           in: query
-          description: Lower boundary in date range query 
+          description: Lower boundary in date range query
           required: false
           schema:
             type: string
         - name: dateTo
           in: query
-          description: Upper boundary in date range query 
+          description: Upper boundary in date range query
           required: false
           schema:
             type: string
@@ -1769,6 +1769,21 @@ paths:
                 properties:
                   response:
                     $ref: '#/components/schemas/ApiResponse'
+                  item:
+                    type: object
+                    properties:
+                      name:
+                        type: string
+                        description: filename of the uploaded content
+                      url:
+                        type: string
+                        description: url of uploaded content
+                      folder:
+                        type: boolean
+                        description: true if content is folder
+                      fileExtension:
+                        type: string
+                        description: file extension of uploaded content
         '400':
           $ref: '#/components/responses/BadRequest'
         '401':
@@ -1843,7 +1858,7 @@ paths:
                   description: Remote repository branch to pull from
                 mergeStrategy:
                   type: string
-                  format: theirs, ours, resolve, recursive, simple two way in core 
+                  format: theirs, ours, resolve, recursive, simple two way in core
                   description: Merge strategy to use when pulling content from remote repository
       responses:
         '200':
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '3.1.1.4'
+  version: '3.1.1.5'
   title: Crafter Studio
   termsOfService: ''
   contact:
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -41,6 +41,8 @@ tags:
   - name: content
     description: Content operations
 
+  - name: configuration
+    description: Configuration operations
 paths:
 
   /groups:
@@ -1228,7 +1230,7 @@ paths:
                   blueprints:
                     type: array
                     items:
-                      $ref: '#/components/schemas/Plugin'
+                      $ref: '#/components/schemas/Blueprint'
         '500':
           $ref: '#/components/responses/InternalServerError'
 
@@ -1315,19 +1317,19 @@ paths:
               type: string
         - name: includeParameters
           in: query
-          description: Include parameters into result set if `true`. Default `false`
+          description: Include parameters into result set if `true`. Default `false` 
           required: false
           schema:
             type: boolean
         - name: dateFrom
           in: query
-          description: Lower boundary in date range query
+          description: Lower boundary in date range query 
           required: false
           schema:
             type: string
         - name: dateTo
           in: query
-          description: Upper boundary in date range query
+          description: Upper boundary in date range query 
           required: false
           schema:
             type: string
@@ -1767,21 +1769,6 @@ paths:
                 properties:
                   response:
                     $ref: '#/components/schemas/ApiResponse'
-                  item:
-                    type: object
-                    properties:
-                      name:
-                        type: string
-                        description: filename of the uploaded content
-                      url:
-                        type: string
-                        description: url of uploaded content
-                      folder:
-                        type: boolean
-                        description: true if content is folder
-                      fileExtension:
-                        type: string
-                        description: file extension of uploaded content
         '400':
           $ref: '#/components/responses/BadRequest'
         '401':
@@ -1856,7 +1843,7 @@ paths:
                   description: Remote repository branch to pull from
                 mergeStrategy:
                   type: string
-                  format: theirs, ours, resolve, recursive, simple two way in core
+                  format: theirs, ours, resolve, recursive, simple two way in core 
                   description: Merge strategy to use when pulling content from remote repository
       responses:
         '200':
@@ -2040,6 +2027,108 @@ paths:
           $ref: '#/components/responses/NotFound'
         '500':
           $ref: '#/components/responses/InternalServerError'
+
+  /configuration/get_configuration:
+    get:
+      tags:
+        - configuration
+      summary: Get configuration content for site and configuration location
+      description: 'Required permission "read_configuration"'
+      operationId: getConfiguration
+      parameters:
+        - name: siteId
+          in: query
+          description: Site ID
+          required: true
+          schema:
+            type: string
+        - name: module
+          in: query
+          description: Module name (e.g. studio, engine)
+          required: true
+          schema:
+            type: string
+        - name: location
+          in: query
+          description: Configuration file location, relative to /config/<module>
+          schema:
+            type: string
+        - name: environment
+          in: query
+          description: environment to use
+          required: false
+          schema:
+            type: string
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+                  content:
+                    type: string
+                    description: configuration file content
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '500':
+          $ref: '#/components/responses/InternalServerError'
+  /configuration/write_configuration:
+    post:
+      tags:
+        - configuration
+      summary: Write configuration content for site
+      description: 'Required permission "write_configuration"'
+      operationId: writeConfiguration
+      requestBody:
+        description: write configuration request body
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                siteId:
+                  type: string
+                  description: site ID
+                module:
+                  type: string
+                  description: Module name (e.g. studio, engine)
+                location:
+                  type: string
+                  description: Configuration file location, relative to /config/<module>
+                environment:
+                  type: string
+                  description: environment to use
+                content:
+                  type: string
+                  description: configuration file content
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '500':
+          $ref: '#/components/responses/InternalServerError'
+
 
   /content/quick_create:
     get:
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -2063,9 +2063,9 @@ paths:
           required: true
           schema:
             type: string
-        - name: location
+        - name: path
           in: query
-          description: Configuration file location, relative to /config/<module>
+          description: Configuration file path, relative to /config/<module>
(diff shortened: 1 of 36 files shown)


src/main/api/2/studio-api-2.yaml
```diff
@@ -2041,6 +2043,108 @@ paths:
...
+          required: true
+          schema:
+            type: string
+        - name: module
+          in: query
+          description: Module name (e.g. studio, engine)
+          required: true
+          schema:
+            type: string
+        - name: location
```
sumerjabri: In the configuration, this is called `path`. Both `path` and `location` are not exactly right, and they're different.

This is a relative path off of `/config/<module>/`.

Just for us to think about this.

src/main/java/org/craftercms/studio/api/v2/service/config/ConfigurationService.java
```diff
@@ -30,4 +35,10 @@
 
     Map<String, List<String>> geRoleMappings(String siteId) throws ConfigurationException;
 
+    String loadConfiguration(String siteId, String module, String location, String environment);
```
sumerjabri: JavaDocs, please.

sumerjabri: getConfigurationAsString

src/main/java/org/craftercms/studio/api/v2/service/config/ConfigurationService.java
```diff
@@ -30,4 +35,10 @@
...
+
+    Document loadConfigurationDocument(String siteId, String module,  String location, String environment)
```
sumerjabri: getConfigurationAsDocument

sumerjabri: Who uses this method? Why do they want the configuration as a DOM? (Let's discuss so we can plan the future where it's all YAML.)

src/main/java/org/craftercms/studio/impl/v2/service/configuration/ConfigurationServiceImpl.java
```diff
@@ -83,13 +110,105 @@ private String getSiteRoleMappingsConfigPath(String siteId) {
...
+            throws DocumentException, IOException {
+        String content = loadEnvironmentConfiguration(siteId, module, location, environment);
+        Document retDocument = null;
+        SAXReader saxReader = new SAXReader();
+        try {
+            saxReader.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+            saxReader.setFeature("http://xml.org/sax/features/external-general-entities", false);
+            saxReader.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
+        }catch (SAXException ex){
+            logger.error("Unable to turn off external entity loading, This could be a security risk.", ex);
```
sumerjabri: small case `this`

src/main/resources/crafter/studio/studio-config.yaml
```diff
@@ -78,8 +78,12 @@ studio.configuration.global.ui.resource.override.path: crafter/studio/ui
 studio.configuration.global.systemSite: studio_root
 # Root location where configuration for all sites is stored inside repository.
 studio.configuration.site.configBasePath: /config/studio
-# Root location where configuration for all sites is stored inside repository.
+# Root location pattern where configuration for all sites is stored inside repository.
+studio.configuration.site.configBasePathPattern: /config/{module}
+# Root location where configuration for all sites is stored inside repository for specified environment.
 studio.configuration.site.multiEnvironment.configBasePath: /config/studio/env/{environment}
```
sumerjabri, line 84: Why not just have one general pattern for all modules. We should follow this pattern across all modules, no?
`/config/{module}/env/{env}/`

src/main/resources/crafter/studio/studio-config.yaml
```diff
@@ -78,8 +78,12 @@ studio.configuration.global.ui.resource.override.path: crafter/studio/ui
...
+# Root location pattern where configuration for all sites is stored inside repository for specified environment.
+studio.configuration.site.multiEnvironment.configBasePathPattern: /config/studio/env/{environment}
```
sumerjabri: See above.

src/main/resources/crafter/studio/studio-config.yaml
```diff
@@ -78,8 +78,12 @@ studio.configuration.global.ui.resource.override.path: crafter/studio/ui
...
 # Location pattern where environment specific configuration for a site is stored
 studio.configuration.site.environment.configBasePath: /config/studio/environment
```
sumerjabri: What's this?

src/main/resources/crafter/studio/studio-config.yaml
```diff
@@ -95,7 +99,7 @@ studio.configuration.site.roleMappingsFileName: role-mappings-config.xml
 # Defines name used for environment specific configuration. It is used for environment overrides in studio. Default value is local.
 studio.configuration.site.environment: local
 # File name where environment specific configuration is stored
-studio.configuration.site.environment.configFileName: environment-config.xml
+studio.configuration.site.environment.configFileName: environment/environment-config.xml
```
sumerjabri, line 100: These are big changes, are you adding handlers in the upgrade manager?

src/main/resources/crafter/studio/studio-config.yaml
```diff
@@ -109,7 +113,7 @@ studio.configuration.site.preview.destroy.context.url: http://localhost:8080/api
 # Location pattern where default dependency resolver specific configuration for is stored
 studio.configuration.default.dependencyResolver.configBasePath: /configuration/dependency
 # File name where default dependency specific configuration is stored
-studio.configuration.default.dependencyResolver.configFileName: resolver-config.xml
+studio.configuration.default.dependencyResolver.configFileName: dependency/resolver-config.xml
```
sumerjabri, line 114: These are big changes, are you adding handlers in the upgrade manager?

src/main/webapp/repo-bootstrap/global/blueprints/1000_website_editorial/config/studio/administration/config-list.xml
```diff
@@ -21,163 +21,189 @@
...
     - samplePath: the path to a sample file to help the user update the file
 -->
 <config>
-    <version>5</version>
+    <version>6</version>
     <files>
         <file>
-            <path>/studio/site-config.xml</path>
+            <module>studio</module>
+            <path>site-config.xml</path>
```
sumerjabri, line 45: This is where we should decide on `path` or `location`. If we change `path` to `location` it'll mean more XSLT updates.

sumerjabri (changes requested):

src/main/resources/crafter/studio/upgrade/3.1.x/3.1.0.29/site/config-list.xslt
```diff
@@ -0,0 +1,116 @@
...
+    </xsl:template>
+
+    <!-- insert line breaks before and after top level comments -->
+    <xsl:template match="/comment()">
+        <xsl:text>&#10;</xsl:text>
+        <xsl:copy-of select="."/>
+        <xsl:text>&#10;</xsl:text>
+    </xsl:template>
+
+    <xsl:template name="search-and-replace">
```
jdrossl, line 39: If this is a regular text operation we have an `FindAndReplaceUpgradeOperation` that could be used to avoid repeating this in all the XSLT files and run it once for all config files that need it

src/main/resources/crafter/studio/upgrade/pipelines.yaml
```diff
@@ -548,6 +557,16 @@ configurations:
           - type: xsltFileUpgrader
             path: *permission-mappings-config
             template: crafter/studio/upgrade/update-version.xslt
+      - currentVersion: 4
+        nextVersion: 5
+        operations:
+          - type: xsltFileUpgrader
+            path: *permission-mappings-config
+            template: crafter/studio/upgrade/3.1.x/3.1.0.29/site/permission-mappings-config.xslt
```
jdrossl, line 565: Maybe we could place these files in a `3.1.0.29/config` folder to make it clear that they will not change site content

jdrossl (changes requested):

src/main/java/org/craftercms/studio/model/rest/WriteConfigurationRequest.java
```diff
@@ -0,0 +1,67 @@
...
+
+package org.craftercms.studio.model.rest;
+
+public class WriteConfigurationRequest {
+
+    private String siteId;
+    private String module;
+    private String location;
+    private String environment;
+    private String content;
```
jdrossl, line 26: This is the whole XML right? Could there be any error when Jackson parses it from JSON?

src/main/api/2/studio-api-2.yaml
dejan-brkic: Agreed. I didn't want to make too many changes in the configuration.

src/main/java/org/craftercms/studio/api/v2/service/config/ConfigurationService.java
dejan-brkic: API v1 services are using DOM to extract properties. I wanted to have SAX parsing in a single class. Future refactoring and possible switching to yaml will be easier this way.

src/main/resources/crafter/studio/studio-config.yaml
dejan-brkic: Small steps. When we fully refactor configuration service, we will have a general pattern

src/main/resources/crafter/studio/studio-config.yaml
dejan-brkic: upgrade manager takes care of all config pattern changes. this one is handled by patterns above (previous comments).

src/main/resources/crafter/studio/studio-config.yaml
dejan-brkic: upgrade manager takes care of all config pattern changes. this one is handled by patterns above (previous comments).

src/main/webapp/repo-bootstrap/global/blueprints/1000_website_editorial/config/studio/administration/config-list.xml
dejan-brkic: I think it might be better to make final decision when we switch to yaml

src/main/resources/crafter/studio/upgrade/3.1.x/3.1.0.29/site/config-list.xslt
dejan-brkic: This XSLT is a complex transformation and `search-and-replace` is part of it. In general use case,  XSLT can be replaced with string + regex operations. Then we don't need XSLT upgrader at all, no?
Personally, XSLT gives more flexibility while writing it. I can produce and test XSLT using online tester tools and when I am happy with the result I can add the final version to upgrade pipeline. I find this process faster and more comfortable than iterating build-start-stop studio to test 'FindAndReplaceUpgradeOperation'

src/main/resources/crafter/studio/upgrade/pipelines.yaml
dejan-brkic: `config` folder is part of the site content. Most of the changes we have in upgrade pipeline are on config files

src/main/api/2/studio-api-2.yaml
sumerjabri: Let's change this to `path.

src/main/java/org/craftercms/studio/model/rest/WriteConfigurationRequest.java
dejan-brkic: So far no errors found.

src/main/webapp/repo-bootstrap/global/blueprints/1000_website_editorial/config/studio/administration/config-list.xml
sumerjabri: I don't want to move forward with two names. Let's change the name to `path`, see my note above.

src/main/api/2/studio-api-2.yaml
dejan-brkic: I need to update UI also

src/main/resources/crafter/studio/upgrade/3.1.x/3.1.0.29/site/config-list.xslt
sumerjabri: find . -name 'xyz.xml'| xargs perl -pi -e "s/before/after/g;" is just as easy to do without build/start/stop.

Having said that, it's fine to do things in XSLT if that works for you. XSLT won't work for everything, however, so we do need our FindAndReplaceUpgradeOperation (and it's faster).

src/main/resources/crafter/studio/upgrade/pipelines.yaml
sumerjabri: `config` is outside of `site`, no?

Or are you saying that anything in the BP is `site`?

dejan-brkic: `config` is outside of `site`, but in terms of structure it is inside site repository. maybe we don't render that content but it belongs to the site

src/main/resources/crafter/studio/upgrade/3.1.x/3.1.0.29/site/config-list.xslt
jdrossl: Yeah, I was just suggesting to do XSLTs just to add the new nodes and then a general FindAndReplace to simplify the templates.

src/main/resources/crafter/studio/upgrade/pipelines.yaml
jdrossl: right, it is part of the repository but it doesn't contain any authored data, that is why I was suggesting to name the folder different

src/main/java/org/craftercms/studio/model/rest/WriteConfigurationRequest.java
jdrossl: Maybe it would be good to also test yaml content as you mentioned it is planned for the future

dejan-brkic: yaml not relevant at the moment

src/main/resources/crafter/studio/upgrade/pipelines.yaml
dejan-brkic: all 28 previous upgrades have `site` and `system` only. just following naming convention already in place

src/main/resources/crafter/studio/upgrade/3.1.x/3.1.0.29/site/config-list.xslt
dejan-brkic: if you review the whole XSLT you will see it adds a new node and replaces the content at same time

src/main/java/org/craftercms/studio/model/rest/WriteConfigurationRequest.java
sumerjabri: I don't understand this part.

sumerjabri: I discussed with Dejan. We should be okay at this time. XML parsing errors will result in `bad request` back to caller.

src/main/resources/crafter/studio/upgrade/pipelines.yaml
sumerjabri: We can separate into content vs config in the future. Right now most upgrades against a site are happening against config.

sumerjabri (approved):

PR #1629, opened 2019-06-14
Add ui tasks to build process

jdrossl:
https://github.com/craftercms/craftercms/issues/3118


pom.xml
```diff
@@ -599,6 +601,67 @@
...
+                    </execution>
+                    <execution>
+                        <id>yarn build sass</id>
+                        <goals>
+                            <goal>yarn</goal>
+                        </goals>
+                        <phase>prepare-package</phase>
+                        <configuration>
+                            <workingDirectory>${studio.ui.path}/ui/scss</workingDirectory>
+                            <arguments>run build</arguments>
```
rart, line 639: No biggie but if this is passed directly on to yarn, the `run` is unnecessary.

PR #1647, opened 2019-06-24
CRAFTERCMS-2364

dejan-brkic:
Added authentication requirements configuration

https://github.com/craftercms/craftercms/issues/2364

### DIFF
--- a/src/main/java/org/craftercms/studio/controller/rest/v2/ExceptionHandlers.java
+++ b/src/main/java/org/craftercms/studio/controller/rest/v2/ExceptionHandlers.java
@@ -39,6 +39,7 @@ import org.craftercms.studio.api.v2.exception.ClusterMemberAlreadyExistsExceptio
 import org.craftercms.studio.api.v2.exception.ClusterMemberNotFoundException;
 import org.craftercms.studio.api.v2.exception.InvalidParametersException;
 import org.craftercms.studio.api.v2.exception.OrganizationNotFoundException;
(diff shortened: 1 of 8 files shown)


src/main/java/org/craftercms/studio/controller/rest/v2/ExceptionHandlers.java
```diff
@@ -203,6 +204,14 @@ public ResponseBody handleCmisPathNotFoundException(HttpServletRequest request,
         return handleExceptionInternal(request, e, response);
     }
 
+    @ExceptionHandler(PasswordRequirementsFailedException.class)
+    @ResponseStatus(HttpStatus.INTERNAL_SERVER_ERROR)
+    public ResponseBody handlePasswordRequirementsFailedException(HttpServletRequest request,
+                                                                  PasswordRequirementsFailedException e) {
+        ApiResponse response = new ApiResponse(ApiResponse.INTERNAL_SYSTEM_FAILURE);
```
sumerjabri: It's not an internal system failure, it's a bad param. Make sure to include a more detailed version of the error in the response codes so the API know what's wrong, and remediation.

sumerjabri (changes requested):

russdanner:
I will confirm this but I will open another ticket to add a user message in the same YAML file for the response.  The error message going back to the user should be able to tell them what the requirements are which must be supplied by the person who wrote the rules.

"User password does not fullfill requirements" is not a user-friendly response.

PR #1655, opened 2019-07-02
CRAFTERCMS-3231

dejan-brkic:
Fixed wrong element

https://github.com/craftercms/craftercms/issues/3231


src/main/resources/crafter/studio/upgrade/3.1.x/3.1.0.34/system/sample-site-config-tools.xslt
```diff
@@ -51,22 +51,6 @@
         </xsl:copy>
     </xsl:template>
 
-    <xsl:template match="config/tools/tool/datasources">
```
sumerjabri, line 54: Why are we removing this?

dejan-brkic: duplicate code. if you check the whole file you will see code block repeated

sumerjabri (approved):

PR #1666, opened 2019-07-12
Fix #3280: User API - usernames with dots are truncated

suelje:
Fix https://github.com/craftercms/craftercms/issues/3280: User API - usernames with dots are truncated

Add a regexp in mapping pattern for /api/2/users/{id}

See https://github.com/spring-projects/spring-framework/issues/10832 or https://github.com/spring-projects/spring-framework/issues/12288

### DIFF
--- a/src/main/resources/crafter/studio/extension/rendering-overlay-context.xml
+++ b/src/main/resources/crafter/studio/extension/rendering-overlay-context.xml
@@ -36,6 +36,10 @@
         <entry key="/*" value-ref="crafter.pageRenderController"/>
     </util:map>
 
+    <mvc:annotation-driven>
+        <mvc:path-matching registered-suffixes-only="true" />
+    </mvc:annotation-driven>
+
     <!-- ////////////////////////////////////// -->
     <!--      Controllers                       -->
     <!-- ////////////////////////////////////// -->


src/main/java/org/craftercms/studio/controller/rest/v2/UsersController.java
```diff
@@ -191,7 +191,7 @@ public ResponseBody deleteUser(
      * @param userId User identifier
      * @return Response containing user
      */
-    @GetMapping("/api/2/users/{id}")
+    @GetMapping("/api/2/users/{id:.+}")
```
jdrossl: What happens if someone calls the endpoint with an extension? for example `../api/2/users/john.doe.json` will the `id` value include the extension?

suelje: Jepp, hmmm you're right. That is an issue: 

`
{
    response: {
        code: 6000,
        message: "User not found",
        remedialAction: "Check if you're using the correct User ID",
        documentationUrl: ""
    }
}`

jdrossl: @suelje maybe you can try adding this configuration and see if it fixes the issue without breaking the extension recognition: `<mvc:path-matching registered-suffixes-only="true"/>`

suelje:
Using `<mvc:path-matching registered-suffixes-only="true"/>` works fine. Thank you. I've updated the Commit.

avasquez614 (approved):

jdrossl (approved):

PR #1672, opened 2019-07-23
#3271

vestrada:
[studio] Default site blueprints should enforce postfixes by default and have all content type fields with postfixes #3271


src/main/webapp/repo-bootstrap/global/blueprints/1000_website_editorial/config/studio/content-types/component/header/form-definition.xml
```diff
@@ -291,7 +291,7 @@
 					<fields>
 				<field>
 					<type>dropdown</type>
-					<id>social_media</id>
+					<id>social_media_o</id>
```
jdrossl: `dropdown` fields should not use `_o`, they should use the same postfixes as the `input`

src/main/webapp/repo-bootstrap/global/blueprints/1000_website_editorial/templates/web/components/header.ftl
```diff
@@ -10,8 +10,8 @@
     </a>
 
     <ul class="icons">
-    <#list contentModel.social_media_links.item as item>
-        <li><a href="${item.url}" class="icon ${item.social_media}"></a></li>
+    <#list contentModel.social_media_links_o.item as item>
+        <li><a href="${item.url_s}" class="icon ${item.social_media_o}"></a></li>
```
jdrossl: See the comment on the definition for `social_media`, because this field is a string it should use `_s`

jdrossl (changes requested):

vestrada:
Done

jdrossl (approved):

PR #1687, opened 2019-07-31
Add new asset controller:

jdrossl:
- Add new controller to serve site files as static resources

https://github.com/craftercms/craftercms/issues/3328


src/main/api/2/studio-api-2.yaml
```diff
@@ -2309,6 +2309,69 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /configuration/extension/file:
```
sumerjabri: perhaps `/plugin` is better because it's not always a configuration nor an extension? (For example, it's a plugin to the form engine, like a picker)

Let's discuss.

src/main/api/2/studio-api-2.yaml
```diff
@@ -2309,6 +2309,69 @@ paths:
...
+    get:
+      tags:
+        - configuration
```
sumerjabri: The same thing, perhaps we change it to `plugin`.

src/main/api/2/studio-api-2.yaml
```diff
@@ -2309,6 +2309,69 @@ paths:
...
+      summary: Get a file for a given extension
+      operationId: getExtensionFile
+      parameters:
+        - name: siteId
```
sumerjabri, line 2321: This means the API is strictly extending sites only. Should we consider making the API more general to address:
- Organization <== we can hide that for now
- Site being optional (so you can have a Main Menu plugin for example

src/main/api/2/studio-api-2.yaml
```diff
@@ -2309,6 +2309,69 @@ paths:
...
+          in: query
+          description: The site ID
+          required: true
+          schema:
+            type: string
+        - name: module
```
sumerjabri: Why is this needed? This is a Studio API only, no?

sumerjabri:
@jdrossl let's meet and discuss this in more detail.

src/main/api/2/studio-api-2.yaml
```diff
@@ -2309,6 +2309,69 @@ paths:
...
+          required: true
+          schema:
+            type: string
+        - name: name
+          in: query
+          description: The extension name (e.g. custom-input)
+          required: true
+          schema:
+            type: string
+        - name: filename
```
sumerjabri, line 2339: Should we also allow the caller to ask for a plugin descriptor where we get the meta-data about it and not just files? Should it be in the same API definition or add another definition?

sumerjabri (changes requested):
Let's discuss.

src/main/api/2/studio-api-2.yaml
sumerjabri: Please make it `/plugin/file`

src/main/api/2/studio-api-2.yaml
sumerjabri: Decision: Keep `siteId` required for now and change that when we're ready for global plugings.

src/main/api/2/studio-api-2.yaml
sumerjabri: Decision: remove this.

src/main/api/2/studio-api-2.yaml
sumerjabri: Decision: Future feature.

src/main/java/org/craftercms/studio/api/v2/service/config/ConfigurationService.java
```diff
@@ -82,6 +82,15 @@ Document getConfigurationAsDocument(String siteId, String module, String path, S
...
+    /**
+     * Get a a file from a plugin
+     * @param siteId the id of site
+     * @param type the type of plugin
+     * @param plugin the name of the plugin
+     * @param filename the path and name of the file
+     * @return the file as a resource
+     * @throws ContentNotFoundException if there is any issue reading the file from the repository
+     */
+    Resource getPluginFile(String siteId, String type, String plugin, String filename)
```
sumerjabri, line 94: Any other exceptions for us to consider?

src/main/java/org/craftercms/studio/api/v1/service/content/ContentService.java
```diff
@@ -77,6 +78,16 @@
...
 
+    /**
+     * Returns content wrapped as a {@link Resource} instance
+     * @param site the site id
+     * @param path the path of the content
+     * @return the resource object
+     * @throws ContentNotFoundException if there is no content at the given path
+     * @since 3.1.1
+     */
+    Resource getContentAsResource(String site, String path) throws ContentNotFoundException;
```
sumerjabri, line 89: Other exceptions?

src/main/java/org/craftercms/studio/controller/rest/v2/ConfigurationController.java
```diff
@@ -17,6 +17,8 @@
 
 package org.craftercms.studio.controller.rest.v2;
 
+import java.io.InputStream;
```
sumerjabri, line 20: Why add this?

src/main/java/org/craftercms/studio/impl/v1/service/configuration/ServicesConfigImpl.java
```diff
@@ -335,6 +344,8 @@ protected SiteConfigTO loadConfiguration(String site) {
              siteConfig.setLastUpdated(ZonedDateTime.now(ZoneOffset.UTC));
 
              loadFacetConfiguration(configNode, siteConfig);
+
+             siteConfig.setPluginFolderPattern(configNode.valueOf("plugin-folder-pattern"));
```
sumerjabri: Please use a constant for the literal string. You can ask Dejan for where to put that.

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/api/v1/service/content/ContentService.java
jdrossl: Since the service just returns the content of the file without any processing the only exception that can be found is that one

src/main/java/org/craftercms/studio/controller/rest/v2/ConfigurationController.java
jdrossl: It was already there, the IDE just moved it to the top of the imports

src/main/java/org/craftercms/studio/api/v2/service/config/ConfigurationService.java
jdrossl: This is the only exception that the service can throw (except for runtime exceptions but those are already handled to return 500)

sumerjabri (approved):

PR #1711, opened 2019-08-19
Add Marketplace support:

jdrossl:
- Add configurable URL in the configuration file
- Add new controller & service for searching plugins

https://github.com/craftercms/craftercms/issues/3343


src/main/api/2/studio-api-2.yaml
```diff
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '3.1.1.15'
+  version: '3.1.1.16'
```
sumerjabri: I think we're on `17` right now.

src/main/api/2/studio-api-2.yaml
```diff
@@ -2721,6 +2723,47 @@ paths:
...
           $ref: '#/components/responses/InternalServerError'
 
+  /marketplace/search:
+    get:
+      tags:
+        - marketplace
+      summary: Search for compatible plugins in the marketplace
+      operationId: searchPlugins
+      parameters:
+        - name: type
```
sumerjabri, line 2733: Can/should we provide examples?

src/main/api/2/studio-api-2.yaml
```diff
@@ -3026,81 +3069,108 @@ components:
...
+              type: string
+              description: Build date of the plugin
+        license:
+          $ref: '#/components/schemas/Link'
+        crafterCmsVersions:
+          type: array
+          description: The supported Crafter CMS versions of the plugin
+          items:
+            $ref: '#/components/schemas/Version'
+        crafterCmsEditions:
```
sumerjabri, line 3147: Should we be returning this to the UI? Or just return Community to Community, and everything to E?
Should we disallow install of E on C in the backend regardless of what the UI lists?
Should the UI explain that E won't work with C?

I feel it best to simply not list E in C, and list all in E.

Let's discuss what's feasible.

sumerjabri (changes requested):

src/main/api/2/studio-api-2.yaml
jdrossl: I agree, we can return only the ones that are compatible, same for the `crafterCmsVersions`. Should we remove the fields from the result to avoid confusion?

sumerjabri: Perhaps the UI should indicate if something is E to the E customers. C customer won't be bothered because all returned is C.

sumerjabri (approved):

PR #1731, opened 2019-09-04
Add WebDAV to API2:

jdrossl:
- Add new controller to expose the existing service
- Small refactor for request constants used by other services

https://github.com/craftercms/craftercms/issues/3420


src/main/api/2/studio-api-2.yaml
```diff
@@ -1325,6 +1327,105 @@ paths:
...
+          schema:
+            type: integer
+            format: int64
+        - name: path
+          in: query
+          description: The path of the directory to list (defaults to the root)
+          required: false
+          schema:
+            type: string
+        - name: type
```
sumerjabri, line 1358: Either switch to MIME-type or spec out the types.

src/main/api/2/studio-api-2.yaml
```diff
@@ -1325,6 +1327,105 @@ paths:
...
+              schema:
+                type: object
+                properties:
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+                  items:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/WebDavItem'
+        '401':
```
sumerjabri, line 1378: Add 404

jdrossl: Actually the WebDAV client never fails, it just returns an empty list

sumerjabri (changes requested):
Please add permissions per our convo.

avasquez614 (approved):

sumerjabri (approved):

PR #1738, opened 2019-09-10
CRAFTERCMS-3422

dejan-brkic:
Completed delete dialog APIs implementation

https://github.com/craftercms/craftercms/issues/3422


src/main/java/org/craftercms/studio/api/v2/dal/DependencyDAO.java
```diff
@@ -55,4 +55,20 @@
...
      */
     List<Map<String, String>> getHardDependenciesForList(Map params);
+
+    /**
+     * Get items depending on given paths
+     *
+     * @param params SQL query parameters
+     * @return List of items depending on given paths
+     */
+    List<String> getDependentItems(Map params);
```
jdrossl: I think it would be a good idea to start using the `@Param` annotation with the actual parameters instead of using a Map for all methods because it helps in reading the code & documenting

jdrossl (approved):

src/main/java/org/craftercms/studio/controller/rest/v2/ContentController.java
```diff
@@ -55,6 +73,37 @@ public ResponseBody listQuickCreateContent(@RequestParam(name = "siteId", requir
         return responseBody;
     }
 
+    @GetMapping("/get_delete_package")
```
jdrossl, line 76: Are these new endpoints already documented in the yaml?

src/main/java/org/craftercms/studio/api/v2/dal/DependencyDAO.java
sumerjabri: @dejan-brkic let's discuss and see if possible.

jdrossl (approved):

sumerjabri (approved):

PR #1745, opened 2019-09-17
Add support for blueprint parameters:

jdrossl:
- Blueprints now include a list of supported parameters
- Create site API will validate the required parameters & replace them in all configuration files

https://github.com/craftercms/craftercms/issues/3002

Depends on https://github.com/craftercms/commons/pull/186

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '3.1.4.2'
+  version: '3.1.4.3'
   title: Crafter Studio
   termsOfService: ''
   contact:
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '3.1.4.1'
+  version: '3.1.4.2'
   title: Crafter Studio
   termsOfService: ''
   contact:
@@ -1333,7 +1333,7 @@ paths:
         - webdav
       summary: Get a list of items from a WebDAV server
       description: 'Required Permission: "webdav_read"'
-      operationId: listItems
+      operationId: listItemsWebdav
       parameters:
         - name: siteId
           in: query
@@ -1386,7 +1386,7 @@ paths:
         - webdav
       summary: Upload a file to a WebDAV server
       description: 'Required Permission: "webdav_write"'
-      operationId: uploadItem
+      operationId: uploadItemWebdav
       requestBody:
         required: true
         content:
@@ -2830,7 +2830,7 @@ paths:
 
   /content/get_delete_package:
     get:
-      tags: 
+      tags:
         - content
       summary: Get a list of delete children (will be deleted) and dependent items (will have broken references) for a given list of items to be deleted
       description: 'Required permission "delete_content"'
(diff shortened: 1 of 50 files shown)


src/main/api/2/studio-api-2.yaml
```diff
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '3.1.4.1'
+  version: '3.1.4.2'
```
dejan-brkic: Please update version with next iteration

src/main/java/org/craftercms/studio/impl/v1/repository/git/StrSubstitutorVisitor.java
```diff
@@ -0,0 +1,90 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.craftercms.studio.impl.v1.repository.git;
```
dejan-brkic: Seems like class does not belong to `git` package. It is not using git operations (plain file system).

src/main/java/org/craftercms/studio/impl/v2/service/site/SitesServiceImpl.java
```diff
@@ -47,6 +49,12 @@ public PluginDescriptor getSiteBlueprintDescriptor(final String id) {
         return sitesServiceInternal.getSiteBlueprintDescriptor(id);
     }
 
+    @Override
+    public void validateBlueprintParameters(final PluginDescriptor descriptor, final Map<String, String> params)
```
dejan-brkic: Why do we want to expose this API and without proper permissions?

dejan-brkic (changes requested):

src/main/java/org/craftercms/studio/impl/v1/repository/git/StrSubstitutorVisitor.java
jdrossl: I'm not sure where to move it, `org.craftercms.studio.impl.v1.repository` or something like `org.craftercms.studio.impl.v1.util`?

src/main/java/org/craftercms/studio/impl/v2/service/site/SitesServiceImpl.java
jdrossl: It needs to be exposed because it is called by `SiteServiceImpl`, I'm not sure what permissions could be applied since it is only used during create site and I think the site service handles permissions (none of the methods in `SitesServiceImpl` has permissions)

dejan-brkic: SiteServiceImpl is V1 service. It will be deprecated at some point. SitesService(Impl) is a public service layer and should have permissions. On the other hand, parameters validation is something that should be performed either at the controller layer (as input validation) or internally within service. I don't see need exposing this API if it will be called only by create site API(s).

src/main/java/org/craftercms/studio/impl/v1/repository/git/StrSubstitutorVisitor.java
dejan-brkic: `util` if it is useful for other services, otherwise `repository`

src/main/java/org/craftercms/studio/impl/v2/service/site/SitesServiceImpl.java
jdrossl: So you think the method should be moved to `SiteServiceImpl` as `protected`? I place it in `SitesServiceImpl ` because all other operations related to blueprint descriptors are there

dejan-brkic (approved):

PR #1756, opened 2019-09-30
CRAFTERCMS-3429

dejan-brkic:
Implemented publishing queue APIs

https://github.com/craftercms/craftercms/issues/3429

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v2/dal/PublishRequest.java
+++ b/src/main/java/org/craftercms/studio/api/v2/dal/PublishRequest.java
@@ -1,0 +1,162 @@
+/*
+ * Copyright (C) 2007-2019 Crafter Software Corporation. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.craftercms.studio.api.v2.dal;
+
+import java.io.Serializable;
+import java.time.ZonedDateTime;
+
+/**
+ * @author Dejan Brkic
+ */
+public class PublishRequest implements Serializable {
+
+    private static final long serialVersionUID = 1287567542272432148L;
+
+    private long id;
+    private String site;
+    private String environment;
+    private String path;
+    private String oldPath;
+    private String user;
+    private ZonedDateTime scheduledDate;
+    private String state;
+    private String action;
+    private String contentTypeClass;
+    private String submissionComment;
+    private String commitId;
+    private String packageId;
+
+    public class State {
+        public final static String READY_FOR_LIVE = "READY_FOR_LIVE";
+        public final static String PROCESSING = "PROCESSING";
+        public final static String COMPLETED = "COMPLETED";
+        public final static String CANCELLED = "CANCELLED";
+        public final static String BLOCKED = "BLOCKED";
+    }
+
+    public class Action {
+        public final static String NEW = "NEW";
+        public final static String UPDATE = "UPDATE";
+        public final static String DELETE = "DELETE";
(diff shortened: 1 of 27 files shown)


src/main/java/org/craftercms/studio/api/v2/dal/PublishRequest.java
```diff
@@ -0,0 +1,162 @@
...
+    public class State {
+        public final static String READY_FOR_LIVE = "READY_FOR_LIVE";
+        public final static String PROCESSING = "PROCESSING";
+        public final static String COMPLETED = "COMPLETED";
+        public final static String CANCELLED = "CANCELLED";
+        public final static String BLOCKED = "BLOCKED";
+    }
+
+    public class Action {
+        public final static String NEW = "NEW";
```
sumerjabri, line 53: Should we name this `CREATE`?

src/main/java/org/craftercms/studio/controller/rest/v2/PublishController.java
```diff
@@ -0,0 +1,100 @@
...
+        result.setTotal(total);
+        result.setOffset(offset);
+        result.setLimit(CollectionUtils.isEmpty(packages) ? 0 : packages.size());
+        result.setResponse(OK);
+        result.setEntities(RESULT_KEY_PACKAGES, packages);
+        responseBody.setResult(result);
+        return responseBody;
+    }
+
+    @GetMapping("/package_details")
```
sumerjabri: Rename to `/package` or `/package/{id}`?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/api/v2/dal/PublishRequest.java
```diff
@@ -0,0 +1,162 @@
...
+    private String user;
+    private ZonedDateTime scheduledDate;
+    private String state;
+    private String action;
+    private String contentTypeClass;
+    private String submissionComment;
+    private String commitId;
+    private String packageId;
+
+    public class State {
```
jdrossl, line 44: I think it would be better to use `enum` instead of classes with only constants

src/main/resources/crafter/studio/upgrade/3.1.x/3.1.4.6/system/global-permission-mappings-config.xslt
```diff
@@ -0,0 +1,50 @@
...
+        </xsl:copy>
+    </xsl:template>
+
+    <!-- insert line breaks before and after top level comments -->
+    <xsl:template match="/comment()">
+        <xsl:text>&#10;</xsl:text><xsl:copy-of select="."/><xsl:text>&#10;</xsl:text>
+    </xsl:template>
+
+
+    <xsl:template match="permissions/role/rule/allowed-permissions">
```
jdrossl, line 39: Can't we use the same XSLT for both files?

src/main/resources/crafter/studio/upgrade/pipelines.yaml
```diff
@@ -528,7 +528,19 @@ pipelines:
               dest: /configuration/samples/sample-targeting-config.xml                                                                                      
           commitDetails: Update sample files
         - type: dbVersionUpgrader    
-
+    - currentVersion: 3.1.4.5
+      nextVersion: 3.1.4.6
```
jdrossl, line 532: It looks like we are missing the version update for the create sql script

jdrossl (changes requested):

src/main/java/org/craftercms/studio/api/v2/dal/PublishRequest.java
dejan-brkic: That is port from v1. Per naming convention established long time ago, it declares the file as new (not existing on the delivery node). If changed it will break DB.

src/main/java/org/craftercms/studio/api/v2/dal/PublishRequest.java
dejan-brkic: Port from v1. Used as constant for DB values, therefore easier for data type conversions to remain as a class of string constant. In terms, of code readability it is similar.

src/main/resources/crafter/studio/upgrade/3.1.x/3.1.4.6/system/global-permission-mappings-config.xslt
dejan-brkic: In general use case, they are addressing different repository types and different pipelines. It turned out that the final regular expression for XSLT matching is the same. I prefer to have it in different files for better readability of what was changed.

sumerjabri (approved):

jdrossl (approved):

PR #1758, opened 2019-10-01
Update for blueprint parameters:

jdrossl:
- Add new fields for description & default value
- Fix issue with order of operations when cloning remote repos

https://github.com/craftercms/craftercms/issues/3486

Depends on https://github.com/craftercms/commons/pull/189


src/main/api/2/studio-api-2.yaml
```diff
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '3.1.4.3'
+  version: '3.1.4.4'
```
dejan-brkic: Please check the latest version of API. I think it is already at 3.1.4.4 if not higher

dejan-brkic (changes requested):

dejan-brkic (approved):

PR #1761, opened 2019-10-02
Update create site:

jdrossl:
- remote name parameter is now optional, defaults to `origin`

https://github.com/craftercms/craftercms/issues/3479

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v2/utils/StudioConfiguration.java
+++ b/src/main/java/org/craftercms/studio/api/v2/utils/StudioConfiguration.java
@@ -34,6 +34,7 @@ public interface StudioConfiguration {
     String SITES_REPOS_PATH = "studio.repo.sitesRepoBasePath";
     String SANDBOX_PATH = "studio.repo.siteSandboxPath";
     String REPO_SANDBOX_BRANCH = "studio.repo.siteSandboxBranch";
+    String REPO_REMOTE_NAME = "studio.repo.siteRemoteName";
     String PUBLISHED_PATH = "studio.repo.sitePublishedPath";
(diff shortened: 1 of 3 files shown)


src/main/java/org/craftercms/studio/api/v2/utils/StudioConfiguration.java
```diff
@@ -34,6 +34,7 @@
     String SITES_REPOS_PATH = "studio.repo.sitesRepoBasePath";
     String SANDBOX_PATH = "studio.repo.siteSandboxPath";
     String REPO_SANDBOX_BRANCH = "studio.repo.siteSandboxBranch";
+    String REPO_REMOTE_NAME = "studio.repo.siteRemoteName";
```
sumerjabri: Add the word DEFAULT.

Maybe: REPO_DEFAULT_REMOTE_NAME

src/main/resources/crafter/studio/studio-config.yaml
```diff
@@ -25,6 +25,8 @@ studio.repo.sitesRepoBasePath: sites
 studio.repo.siteSandboxPath: sandbox
 # Sandbox git repository branch for every site
 studio.repo.siteSandboxBranch: master
+# Name used for the git remote for remote sites
+studio.repo.siteRemoteName: origin
```
sumerjabri: studio.repo.defaultRemoteName

sumerjabri (changes requested):
Should we also update the docs to indicate that this param is now optional?

jdrossl:
PR for the doc https://github.com/craftercms/docs/pull/849

sumerjabri (approved):

PR #1777, opened 2019-10-11
Remove old templates from sites

jdrossl:
https://github.com/craftercms/craftercms/issues/3513

### DIFF
--- a/src/main/resources/crafter/studio/upgrade/pipelines.yaml
+++ b/src/main/resources/crafter/studio/upgrade/pipelines.yaml
@@ -721,6 +721,20 @@ pipelines:
           commitDetails: Add default gitignore file
         - type: versionFileUpgrader
 
+    - currentVersion: 3.1.4.2
+      nextVersion: 3.1.4.3
+      operations:
+        - type: deleteUpgrader
+          paths:
+            - /templates/system/common/components-support.ftl
+            - /templates/system/common/crafter-support.ftl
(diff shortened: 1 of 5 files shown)


src/main/resources/crafter/studio/upgrade/pipelines.yaml
```diff
@@ -721,6 +721,20 @@ pipelines:
           commitDetails: Add default gitignore file
         - type: versionFileUpgrader
 
+    - currentVersion: 3.1.4.2
+      nextVersion: 3.1.4.3
```
dejan-brkic, line 725: This version number really looks out of date. I see in git log as latest 3.1.4.9

dejan-brkic (changes requested):

src/main/resources/crafter/studio/upgrade/pipelines.yaml
jdrossl: This is for the `site` pipeline, not `system` https://github.com/craftercms/studio/blob/develop/src/main/resources/crafter/studio/upgrade/pipelines.yaml#L716

dejan-brkic: Yes, I noticed. When I was looking for DB version update noticed it is in site pipeline

dejan-brkic (approved):

PR #1794, opened 2019-10-21
Fix for parameters in orphan sites:

jdrossl:
- Include all files in the initial commit for parameter validation

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepositoryHelper.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepositoryHelper.java
@@ -1037,6 +1037,9 @@ public class GitContentRepositoryHelper {
             logger.debug("Soft reset to commit empty repo");
             git.reset().call();
 
+            // Include all files in the first commit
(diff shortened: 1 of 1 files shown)


dejan-brkic (changes requested):
It will cause https://github.com/craftercms/craftercms/issues/3549.
We need to have the empty repository commit first.
Please find another way to resolve your issue.

src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepositoryHelper.java
```diff
@@ -1037,6 +1037,9 @@ private void makeRepoOrphan(Repository repository, String site) throws IOExcepti
             logger.debug("Soft reset to commit empty repo");
             git.reset().call();
 
+            // Include all files in the first commit
+            git.add().addFilepattern(GIT_COMMIT_ALL_ITEMS).call();
```
dejan-brkic, line 1041: It will cause craftercms/craftercms#3549.
We need to have the empty repository commit first.
Please find another way to resolve your issue.

dejan-brkic (changes requested):

PR #1796, opened 2019-10-22
New create site API for Marketplace:

jdrossl:
- Marketplace service handles validations and triggers the create

Depends on https://github.com/craftercms/commons/pull/192

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '3.1.4.6'
+  version: '3.1.4.7'
   title: Crafter Studio
   termsOfService: ''
   contact:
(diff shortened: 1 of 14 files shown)


src/main/api/2/studio-api-2.yaml
```diff
@@ -3025,6 +3025,35 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /marketplace/create_site:
```
sumerjabri: Why not `/site/create_site_from_marketplace`?

`/site/` becomes the new root for API2 site operations.

My issue with `/marketplace` is that it's misleading thinking this is the marketplace of Studio when it's just create site API specialized.

src/main/java/org/craftercms/studio/impl/v2/service/marketplace/MarketplaceServiceImpl.java
```diff
@@ -149,4 +181,52 @@ protected void validate() throws MarketplaceException {
         }
     }
 
+    public Map<String, Object> getDescriptor(String id, Version version) throws MarketplaceException,
```
dejan-brkic: Public methods of v2 services should be protected with permissions

src/main/java/org/craftercms/studio/impl/v2/service/marketplace/MarketplaceServiceImpl.java
```diff
@@ -149,4 +181,52 @@ protected void validate() throws MarketplaceException {
...
+        } catch (ResourceAccessException e) {
+            throw new MarketplaceUnreachableException(url, e);
+        } catch (IllegalArgumentException e) {
+            throw new BlueprintNotFoundException(String.format("Blueprint not found in the Marketplace: %s %s.%s.%s",
+             id, version.getMajor(), version.getMinor(), version.getPatch()));
+        }
+    }
+
+    @Override
+    public void createSite(CreateSiteRequest request) throws RemoteRepositoryNotFoundException,
```
dejan-brkic, line 68: Public methods of v2 services should be protected with permissions

dejan-brkic (changes requested):

sumerjabri (changes requested):

src/main/api/2/studio-api-2.yaml
jdrossl: I added it there to call directly the marketplace service and avoid having to change the other service, my idea was to wrap the site service so they continue to be independent (mp can change without changing site service and the other way around)

jdrossl: If you are only worried about the URL then we can move it to the `SitesController` and still call the MP service. Let me know if that works or if you want to talk about it

dejan-brkic (approved):

sumerjabri (approved):

PR #1833, opened 2019-11-13
[3671] Feature drag drop fails on editorial home page

jvega190:
https://github.com/craftercms/craftercms/issues/3671


rart (approved):

russdanner (approved):
This looks good

PR #1851, opened 2019-12-16
Add support for encrypted configuration properties

jdrossl:
craftersoftware/craftercms#18

Depends on craftercms/commons#197

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '3.1.4.8'
+  version: '3.1.5.1'
   title: Crafter Studio
   termsOfService: ''
   contact:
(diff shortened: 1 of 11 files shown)


avasquez614 (approved):

src/main/api/2/studio-api-2.yaml
```diff
@@ -3215,6 +3217,37 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /security/encrypt:
+    get:
+      tags:
+        - security
+      summary: Encrypt a text value
+      operationId: encrypt
+      parameters:
```
avasquez614, line 3226: Don't we need a description with the required permission, like `description: 'Required permission "read_configuration"'`?

avasquez614 (changes requested):

src/main/api/2/studio-api-2.yaml
jdrossl: Right now it doesn't require any special permission, just to be authenticated

avasquez614 (approved):

PR #1869, opened 2020-01-27
Added FileEncodingConvertingAssetProcessor

avasquez614:
Ticket craftercms/craftercms#3743


sumerjabri:
Rejecting this approach. Will support via APIs solely.

PR #1881, opened 2020-02-10
[52-e] [studio] When SAML SSO session expires by timeout validate-session returns wrong response

dejan-brkic:
Change validate session API response on session expired

https://github.com/craftersoftware/craftercms/issues/52

### DIFF
--- a/src/main/resources/crafter/studio/studio-config.yaml
+++ b/src/main/resources/crafter/studio/studio-config.yaml
@@ -209,7 +209,7 @@ studio.security.ignoreRenewTokenUrls: /api/1/services/api/1/security/validate-se
 # Public URLs
 studio.security.publicUrls: >
   /api/1/services/api/1/server/get-available-languages.json,/api/1/services/api/1/server/get-ui-resource-override.json,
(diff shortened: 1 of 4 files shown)


src/main/webapp/default-site/scripts/rest/api/1/security/validate-session.get.groovy
```diff
@@ -16,18 +15,24 @@
...
+        result.active = true
+        result.user = UserServices.getCurrentUser(context)
         response.setStatus(200)
     } else {
-        result.message = "Unauthorized"
-        response.setStatus(401)
+        result.message = null
+        result.active = false
+        result.user = null
+        response.setStatus(200)
```
avasquez614, line 35: For consistency, I would say response status should be 401. Is there any reason why now we're sending 200?

avasquez614 (changes requested):

src/main/webapp/default-site/scripts/rest/api/1/security/validate-session.get.groovy
dejan-brkic: The response format is changed as a new feature is added in UI to support SAML session expired

avasquez614 (approved):

PR #1882, opened 2020-02-12
modify order of getAllPublishedEnvironments

phuongnq:
When we have both staging and live, the default order of environment list should be `staging` -> `live`.
`HashSet` cannot guarantee this. `LinkedHashSet` guarantee adding order.


russdanner (approved):
Simple enough change IMO.
I agree on the UX

dejan-brkic (approved):

PR #1884, opened 2020-02-13
Site Plugins:

jdrossl:
- Add new API for getting & installing plugins in sites

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '4.0.0.3'
+  version: '4.0.0.4'
   title: Crafter Studio
   termsOfService: ''
   contact:
@@ -5675,9 +5675,6 @@ components:
         path:
           type: string
           description: The path where the file was copied during installation
-        sha512:
-          type: string
-          description: The SHA-512 checksum of the file during installation
 
     SearchParams:
       type: object
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -3945,6 +3945,88 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /marketplace/installed:
+    get:
+      tags:
+        - marketplace
+      summary: Get the list of plugins installed in the given site
+      operationId: getInstalledPlugins
+      parameters:
+        - name: siteId
+          in: query
+          description: site ID
+          required: true
+          schema:
+            type: string
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+                  plugins:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/PluginRecord'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '500':
+          $ref: '#/components/responses/InternalServerError'
+
+  /marketplace/install:
+    post:
+      tags:
+        - marketplace
+      summary: Install a plugin in the given site
+      operationId: installPlugin
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                siteId:
+                  description: The id of the site
+                  type: string
+                pluginId:
+                  description: The id of the plugin
+                  type: string
+                pluginVersion:
+                  $ref: '#/components/schemas/Version'
+              required:
+                - siteId
+                - pluginId
+                - pluginVersion
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '409':
+          $ref: '#/components/responses/Conflict'
+        '500':
+          $ref: '#/components/responses/InternalServerError'
+
   /publish/packages:
     get:
       tags:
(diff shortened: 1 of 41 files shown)


src/main/api/2/studio-api-2.yaml
```diff
@@ -3096,6 +3096,84 @@ paths:
...
+                    $ref: '#/components/schemas/ApiResponse'
+                  plugins:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/PluginRecord'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '500':
```
sumerjabri: Add 403?

src/main/api/2/studio-api-2.yaml
```diff
@@ -3096,6 +3096,84 @@ paths:
...
+              schema:
+                type: object
+                properties:
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '409':
```
sumerjabri: Add 403?

src/main/api/2/studio-api-2.yaml
```diff
@@ -3777,6 +3855,37 @@ components:
...
+          type: string
+          format: date-time
+          description: The date when the plugin was installed
+        files:
+          type: array
+          description: The list of installed files
+          items:
+            $ref: '#/components/schemas/FileRecord'
+
+    FileRecord:
```
sumerjabri, line 5674: Let's discuss this part over a call.

src/main/java/org/craftercms/studio/model/rest/ApiResponse.java
```diff
@@ -120,6 +120,9 @@
         "Marketplace server is unreachable", "Check the configuration to make sure the Marketplace URL is correct",
         StringUtils.EMPTY);
 
+    public static final ApiResponse PLUGIN_ALREADY_INSTALLED =
+            new ApiResponse(40002, "Plugin is already installed", StringUtils.EMPTY, StringUtils.EMPTY);
+
     public static final ApiResponse CMIS_UNREACHABLE =
             new ApiResponse(50000, "CMIS server is unreachable", "Advise the user that the CMIS server is not " +
```
sumerjabri: Was this added to the list of error codes in the OAS docs?

src/main/resources/crafter/studio/studio-config.yaml
```diff
@@ -787,3 +787,24 @@ studio.serverless.delivery.deployer.target.template.params:
...
+studio.marketplace.plugin.installable:
+  - engine
+# The name of the sub-folder to copy plugin files
+studio.marketplace.plugin.folder: plugins
+# The mapping of folders to copy files from the plugin folder to the site
+studio.marketplace.plugin.mapping:
+  - templates: templates
+  - static-assets: static-assets
+  - scripts/components: scripts/components
+  - scripts/filters: scripts/filters
```
sumerjabri: Let's discuss.

sumerjabri (changes requested):

src/main/api/2/studio-api-2.yaml
jdrossl: Right now there are no permissions defined for these new operations, any authenticated user can execute them

src/main/java/org/craftercms/studio/impl/v2/service/marketplace/MarketplaceServiceImpl.java
```diff
@@ -71,4 +75,17 @@ public void createSite(CreateSiteRequest request) throws RemoteRepositoryNotFoun
...
     }
 
+    @Override
+    @ValidateParams
+    public List<PluginRecord> getInstalledPlugins(@ValidateStringParam(name = "siteId") String siteId)
+            throws MarketplaceException {
+        return marketplaceServiceInternal.getInstalledPlugins(siteId);
+    }
+
+    //TODO: Add specific permissions?
```
sumerjabri: Permission is "site member"

rart (commented):
Looks good.

The only feedback I’ve got is if we could, for the `/installed` endpoint, use the same response model as marketplace search ([MarketplacePlugin](https://app.swaggerhub.com/apis/craftercms/studio/4.0.0.1#/MarketplacePlugin), [GitMarketplacePlugin](https://app.swaggerhub.com/apis/craftercms/studio/4.0.0.1#/GitMarketplacePlugin)) which includes media, compatibility, and other plugin details. This would allow for a more visual and rich listing and details UI. The properties of the new model are of use too, so maybe combine the models, or nest them, or be able to relate them somehow (e.g. pluginRecord.pluginDescriptor?)

Question: will local site plugins – i.e. plug-ins that were developed and deployed with the site, without coming from the marketplace – also be indexed, returned and managed by these APIs? An additional property to flag whether a record came from marketplace or is “local” would be appropriate.

jdrossl:
@rart
Local plugins developed as part of the site will not be returned by this API, only plugins installed from the marketplace.

About merging the response I don't think it's a good idea for now, but we can consider it later. The reason is that this API returns data from a file stored in the site, if there are 10 plugins installed we would need to query the MP for each one of the descriptors.

rart:
Does the info of the MarketplacePlugin models come from the yaml file or it comes from some sort of database? Installed plugins have their yaml file, right? Could we extract directly from the yaml?

jdrossl:
We are not storing the descriptors right now

sumerjabri (changes requested):
Roles allowed to install plugins:
* Developers
* Admins

Roles allowed to list plugins:
* Everyone in the site

Switch 4.0.0.1 to 4.0.0.4 for API OAS, UM, and other places.

Remove sha512 from FileRecord in OAS

src/main/api/2/studio-api-2.yaml
```diff
@@ -3945,6 +3945,88 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /marketplace/installed:
+    get:
+      tags:
+        - marketplace
+      summary: Get the list of plugins installed in the given site
+      operationId: getInstalledPlugins
```
sumerjabri, line 3954: Please add required permission in the spec, like:

```
description: 'Required Permission: "XXYYZZ"'
```

sumerjabri (changes requested):

sumerjabri (approved):

PR #1885, opened 2020-02-14
[3816] [studio] Validate session not added to global repo overrides as public url

dejan-brkic:
Added missing configuration

https://github.com/craftercms/craftercms/issues/3816

### DIFF
--- a/src/main/resources/crafter/studio/upgrade/pipelines.yaml
+++ b/src/main/resources/crafter/studio/upgrade/pipelines.yaml
@@ -726,6 +726,17 @@ pipelines:
               dest: configuration/samples/sample-studio-config-override.yaml
           commitDetails: Update sample for global config override file
         - type: dbVersionUpgrader
+    - currentVersion: 3.1.5.5
+      nextVersion: 3.1.5.6
+      operations:
+        - type: findAndReplaceUpgrader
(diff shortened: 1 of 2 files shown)


avasquez614 (approved):

src/main/resources/crafter/studio/upgrade/pipelines.yaml
```diff
@@ -726,6 +726,17 @@ pipelines:
               dest: configuration/samples/sample-studio-config-override.yaml
           commitDetails: Update sample for global config override file
         - type: dbVersionUpgrader
+    - currentVersion: 3.1.5.5
+      nextVersion: 3.1.5.6
+      operations:
+        - type: findAndReplaceUpgrader
+          includedPaths:
+            - configuration/studio-config-override.yaml
+          pattern: "studio.security.publicUrls: >\\r*\\n*\\s+"
```
jdrossl, line 735: @dejan-brkic what happens if the url is already there, wouldn't it repeat it?

dejan-brkic: this will run only once (for transition from 3.1.5.5 to 3.1.5.6) if it is repeated, it is list of urls so it is only redundant

PR #1902, opened 2020-03-03
Add Blob Storage:

jdrossl:
- Proxy operations to read/write/publish content to the repository
- Support for AWS S3 Blob Store

https://github.com/craftercms/craftercms/issues/3744

Depends on https://github.com/craftercms/commons/pull/207, https://github.com/craftercms/engine/pull/353


src/main/java/org/craftercms/studio/impl/v2/repository/blob/BlobAwareContentRepository.java
```diff
@@ -0,0 +1,540 @@
...
+        try {
+            if (!isFolder(path)) {
+                BlobStore store = getRemoteStore(site, path);
+                if (store != null) {
+                    return store.contentExists(site, path);
+                }
+            }
+            return localRepositoryV1.contentExists(site, path);
+        } catch (Exception e) {
+            return false;
```
avasquez614, line 136: Not logging exceptions?

src/main/java/org/craftercms/studio/impl/v2/repository/blob/BlobStoreFactoryImpl.java
```diff
@@ -0,0 +1,123 @@
...
+    public void setContentRepository(ContentRepository contentRepository) {
+        this.contentRepository = contentRepository;
+    }
+
+    public void setConfigurationReader(EncryptionAwareConfigurationReader configurationReader) {
+        this.configurationReader = configurationReader;
+    }
+
+    @Override
+    public BlobStore getByPaths(String site, String[] paths) throws ServiceLayerException, ConfigurationException {
```
avasquez614: Why not use String... (varargs)?

avasquez614 (changes requested):

dejan-brkic (approved):

src/main/java/org/craftercms/studio/impl/v2/repository/blob/BlobAwareContentRepository.java
avasquez614: This is still not resolved. Can you explain why are you not logging the exception?

avasquez614 (approved):

PR #1927, opened 2020-03-18
[3883] Create site dialog updates - Update in marketplace search params (docs)

jvega190:
craftercms/craftercms#3883
3.1.7 Ticket - Update in marketplace search params (docs)


sumerjabri (approved):

src/main/api/2/studio-api-2.yaml
```diff
@@ -3315,6 +3315,11 @@ paths:
           schema:
             type: integer
             format: int64
+        - name: showIncompatible
+          in: query
+          description: Include incompatible plugins (default to false)
```
rart: defaults* to false

Is there no way of stating it's optionality and default value through actual definition arguments?

jdrossl: yes, by default all parameters are optional (`required: false` is not needed), and it is possible to use `default: false`

rart (approved):

PR #1942, opened 2020-04-03
[3943] [studio] Asset processing field keep original should be optional

dejan-brkic:
fixed bug

https://github.com/craftercms/craftercms/issues/3943


russdanner:
Confirmed

PR #1952, opened 2020-04-15
Use CRAFTER_ENVIRONMENT to set active env

alhambrav:
Use CRAFTER_ENVIRONMENT to set active env

### DIFF
--- a/src/main/resources/crafter/studio/studio-config.yaml
+++ b/src/main/resources/crafter/studio/studio-config.yaml
@@ -587,6 +587,8 @@ studio.cors.credentials: true
 # Value for the Access-Control-Max-Age header
 studio.cors.maxage: -1
 
+# The active environment for multi environment configuration, e.g. qa, prod, dev
+studio.configuration.environment.active: ${env:CRAFTER_ENVIRONMENT}
 
 ################################################################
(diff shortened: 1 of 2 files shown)


src/main/resources/crafter/studio/studio-config.yaml
```diff
@@ -587,6 +587,8 @@ studio.cors.credentials: true
 # Value for the Access-Control-Max-Age header
 studio.cors.maxage: -1
 
+# The active environment for multi environment configuration, e.g. qa, prod, dev
+studio.configuration.environment.active: ${env:CRAFTER_ENVIRONMENT}
```
jdrossl: This should go in the override in craftercms instead of this repo

jdrossl (changes requested):

src/main/resources/crafter/studio/studio-config.yaml
sumerjabri: Should this have something basic/foundational that's overridden by the one in `craftercms`?

jdrossl: It's never been there

jdrossl (approved):

PR #1967, opened 2020-04-23
[studio] EnvironmentOverrides updates

rart:
Removes the no-site-set redirect 
Removes service invocations that tried to read things from environment-config.xml, which is now removed by upgrade manager

https://github.com/craftercms/craftercms/issues/3982
https://github.com/craftercms/craftercms/issues/3889

For more comfortable/easier review of the files tab, [click here](https://github.com/craftercms/studio/pull/1967/files?diff=unified&w=1)

### DIFF
--- a/src/main/webapp/default-site/scripts/libs/EnvironmentOverrides.groovy
+++ b/src/main/webapp/default-site/scripts/libs/EnvironmentOverrides.groovy
@@ -16,95 +16,108 @@
 
 package scripts.libs
 
-import scripts.api.SiteServices;
+import org.craftercms.studio.api.v1.log.LoggerFactory
 import scripts.api.SecurityServices
+import scripts.api.SiteServices
 
-import static org.craftercms.studio.api.v2.utils.StudioConfiguration.SECURITY_PASSWORD_REQUIREMENTS_VALIDATION_REGEX;
-import static org.craftercms.studio.api.v2.utils.StudioConfiguration.SECURITY_TYPE;
+import static org.craftercms.studio.api.v2.utils.StudioConfiguration.SECURITY_PASSWORD_REQUIREMENTS_VALIDATION_REGEX
 
 class EnvironmentOverrides {
-    static SITE_SERVICES_BEAN = "cstudioSiteServiceSimple"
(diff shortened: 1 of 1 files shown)


rart:
@jdrossl I'm trying to remove the need for a site to be already set for internal pages.

If there's no site set, I'm grabbing the first one and setting it. This works, however there's an odd error that happens anyway. It's odd because it says something about `environment-config` which was removed so maybe there's still something somewhere. 

FYI The upgrade manager should have already ran over all my sites.

```
[WARN] 2020-04-23T14:00:05,025 [http-nio-8080-exec-9] [git.GitContentRepositoryHelper] | Couldnt get the sandbox repository for site: UNSET 
[WARN] 2020-04-23T14:00:05,026 [http-nio-8080-exec-9] [git.GitContentRepositoryHelper] | Couldnt get the sandbox repository for site: UNSET 
[WARN] 2020-04-23T14:00:05,026 [http-nio-8080-exec-9] [git.GitContentRepositoryHelper] | Couldnt get the sandbox repository for site: UNSET 
[WARN] 2020-04-23T14:00:05,027 [http-nio-8080-exec-9] [git.GitContentRepositoryHelper] | Couldnt get the sandbox repository for site: UNSET 
[ERROR] 2020-04-23T14:00:05,027 [http-nio-8080-exec-9] [blob.BlobAwareContentRepository] | Error getting content /config/studio/environment/environment-config.xml in site UNSET 
org.craftercms.studio.api.v1.exception.ContentNotFoundException: Repository not found for site UNSET
	at org.craftercms.studio.impl.v1.repository.git.GitContentRepository.getContent(GitContentRepository.java:242) ~[classes/:3.2.0-SNAPSHOT]
	at org.craftercms.studio.impl.v2.repository.blob.BlobAwareContentRepository.getContent(BlobAwareContentRepository.java:162) [classes/:3.2.0-SNAPSHOT]
	at ...
```

jdrossl:
@rart the upgrade manager runs during startup for all existing sites but also after you create a new one just for it. According to the messages `UNSET` is being use as the id of the site, that is causing all those errors

dejan-brkic (approved):

PR #1982, opened 2020-05-06
Fix overwritten changes

jdrossl:


sumerjabri:
@dejan-brkic see if this overlaps with your current work. (Removal of env-conf). We must keep the part that adds the proxy.

jdrossl:
@dejan-brkic if you are working on `develop` this can be merged and later you can apply the same changes in this branch

dejan-brkic (approved):

PR #1985, opened 2020-05-11
[studio] Better UI configuration support

rart:
There's a number of config files which sole purpose is to configure the UI. They are all currently XML and the UI needs to incur in parsing to get it to the final consumption data structure. 

For example the `/preview-tools/panel`: The system makes no use of this config. It only stores and serves to the UI when requested. The UI retrieve parses to transform the XML verbosity and structure into a consumable and efficient format.

The backend is actually agnostic to what type of file gets requested so if one or more config files were to be JSON, there would be no issue in requesting them. However, the UI doesn't know which files would be JSON or XML. This would require 2 requests to fetch the config file.

This PR enables the UI to call the service with a file without the extension and the backend would attempt to first find an XML and if not present find the JSON file.

The UI supports this duality already for the preview tools and the targeting panel. We would only need this change to allow for the backend-side smart resolution of the file.

The change is 100% backward compatible.

### DIFF
--- a/src/main/java/org/craftercms/studio/controller/rest/v2/ConfigurationController.java
+++ b/src/main/java/org/craftercms/studio/controller/rest/v2/ConfigurationController.java
@@ -47,15 +47,29 @@ public class ConfigurationController {
     private StudioConfiguration studioConfiguration;
 
     @GetMapping("/get_configuration")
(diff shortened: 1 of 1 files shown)


rart:
@dejan-brkic any thoughts? (can't add you as reviewer by myself)

sumerjabri:
We have a different approach in mind.

PR #2026, opened 2020-06-02
Improvement for search service:

jdrossl:
- Add optional parameter to filter paths using regexes

https://github.com/craftercms/craftercms/issues/4047

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '3.1.8.1'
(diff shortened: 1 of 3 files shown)


src/main/api/2/studio-api-2.yaml
```diff
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '3.1.7.3'
+  version: '3.1.8'
```
sumerjabri: Please update to 3.1.8.2

sumerjabri (changes requested):

sumerjabri (approved):

PR #2038, opened 2020-06-18
Add translation config API

jdrossl:
https://github.com/craftercms/craftercms/issues/4019


sumerjabri (approved):

src/main/api/2/studio-api-2.yaml
```diff
@@ -4085,6 +4085,39 @@ paths:
 #        '500':
 #          $ref: '#/components/responses/InternalServerError'
 
+
+  /translation/config:
```
rart: Sorry, just noticed this: perhaps `get_config` would be semantically better (or fetch_config, retrieve_config or similar). The name config sounds like is a verb or so. Makes me think this is the service to set a config and possibly would be a post.

`get_config` would also better match the operation ID and other API2 items (e.g. content/get_delete_package).

I do like the omission of "translation" within the operation name as we're already in that namespace (i.e. get_config instead of get_translation_config)

rart (changes requested):

src/main/api/2/studio-api-2.yaml
```diff
@@ -4085,6 +4085,39 @@ paths:
...
+    get:
+      tags:
+        - translation
```
dejan-brkic: This is a translation configuration API, so it should be under configuration space, no ?

src/main/java/org/craftercms/studio/api/v1/service/configuration/ServicesConfig.java
```diff
@@ -16,11 +16,7 @@
 package org.craftercms.studio.api.v1.service.configuration;
 
 
-import org.craftercms.studio.api.v1.to.ContentTypeConfigTO;
-import org.craftercms.studio.api.v1.to.CopyDependencyConfigTO;
-import org.craftercms.studio.api.v1.to.DeleteDependencyConfigTO;
-import org.craftercms.studio.api.v1.to.DmFolderConfigTO;
-import org.craftercms.studio.api.v1.to.FacetTO;
+import org.craftercms.studio.api.v1.to.*;
```
dejan-brkic: AFAIK our code convention says no * imports

src/main/java/org/craftercms/studio/api/v1/to/TranslationConfigTo.java
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.craftercms.studio.api.v1.to;
```
dejan-brkic: Why V1 package space if API is new and V2 ?

src/main/java/org/craftercms/studio/api/v2/service/translation/TranslationService.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.craftercms.studio.api.v1.to.TranslationConfigTo;
+
+/**
+ * Defines all translation related operations
+ *
+ * @author joseross
+ * @since 3.2.0
+ */
+public interface TranslationService {
```
dejan-brkic: Configuration service vs Translation service ?

src/main/java/org/craftercms/studio/controller/rest/v2/TranslationController.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Rest controller to handle all translation related operations
+ *
+ * @author joseross
+ * @since 3.2.0
+ */
+@RestController
+@RequestMapping("/api/2/translation")
+public class TranslationController {
```
dejan-brkic: Again Configuration space vs Translations space

src/main/webapp/repo-bootstrap/global/blueprints/1000_website_editorial/config/studio/site-config.xml
```diff
@@ -69,6 +69,16 @@
         </ignore-postfix-fields>
     </form-engine>
 
+    <!-- Site Translation -->
```
dejan-brkic: Why adding configuration as commented?

src/main/webapp/repo-bootstrap/global/blueprints/2000_headless_store/config/studio/site-config.xml
```diff
@@ -69,6 +69,16 @@
         </ignore-postfix-fields>
     </form-engine>
 
+    <!-- Site Translation -->
+    <!--
```
dejan-brkic: Why adding configuration as commented?

src/main/webapp/repo-bootstrap/global/blueprints/4000_empty/config/studio/site-config.xml
```diff
@@ -69,6 +69,16 @@
         </ignore-postfix-fields>
     </form-engine>
 
+    <!-- Site Translation -->
+    <!--
```
dejan-brkic: Why adding configuration as commented?

src/main/webapp/repo-bootstrap/global/blueprints/5000_headless_blog/config/studio/site-config.xml
```diff
@@ -69,6 +69,16 @@
         </ignore-postfix-fields>
     </form-engine>
 
+    <!-- Site Translation -->
+    <!--
```
dejan-brkic: Why adding configuration as commented?

src/main/webapp/repo-bootstrap/global/configuration/samples/sample-site-config.xml
```diff
@@ -65,6 +65,16 @@
         </ignore-postfix-fields>
     </form-engine>
 
+    <!-- Site Translation -->
+    <translation>
```
dejan-brkic: And here it is not commented?

dejan-brkic (changes requested):

src/main/api/2/studio-api-2.yaml
jdrossl: Sure I'll update to follow the existing pattern, but I think we should consider moving towards a more standard REST API where the URL identifies a resource and the operation is defined by the HTTP method:

```
GET  /translation/config -> retrieves the config
POST /translation/config -> updates the config
```

src/main/api/2/studio-api-2.yaml
jdrossl: @sumerjabri what do you think?

src/main/java/org/craftercms/studio/api/v1/service/configuration/ServicesConfig.java
jdrossl: I didn't knew that, will fix it

src/main/java/org/craftercms/studio/api/v1/to/TranslationConfigTo.java
jdrossl: Will fix it

src/main/webapp/repo-bootstrap/global/blueprints/1000_website_editorial/config/studio/site-config.xml
jdrossl: According to the ticket it should be added as commented for backwards compatibility

src/main/webapp/repo-bootstrap/global/configuration/samples/sample-site-config.xml
jdrossl: @sumerjabri what do you think?

src/main/api/2/studio-api-2.yaml
rart: I've always been in favour of that too. 

I believe at the early stages of API2 that was considered; don't remember the reason why we ended up steering away from that.

src/main/api/2/studio-api-2.yaml
jdrossl: After reviewing the `ConfigurationController` (namespace `/api/2/configuration`) it is a generic read/write controller, this one just reads a specific section of `site-config.xml` so I think they should be separated

dejan-brkic: As our goal/target is to have all configuration in a single file at some point, small steps towards it are to keep services in a single namespace/controller

src/main/webapp/repo-bootstrap/global/blueprints/1000_website_editorial/config/studio/site-config.xml
dejan-brkic: It is a new feature. There is no backward compatibility. Old code does not know about that config so it will never read it if it exists. Only a new code needs to handle config if it does not exist. The config file should have it or not have it. Why adding a new config as a comment? There is a sample file where you added it as an example of config.

jdrossl: I agree, updating the sample should be enough, what do you think @sumerjabri?

sumerjabri: I want the block added and commented out. This way existing sites can uncomment and use the new feature.

src/main/api/2/studio-api-2.yaml
sumerjabri: Let's use a new API as `/api/2/configuration/{siteId}/{module}/{section}/{key}/{value}` with GET/POST.
Params:
module (in), key (in, optional), JSON (in/out).

You can read a section and write a whole section or a parameter in a section.

Example:

`GET /api/2/configuration/{siteId}/studio/translation` -> returns the translation section
`GET /api/2/configuration/{siteId}/studio/translation/defaultLocale` -> returns the default locale for this site
`POST /api/2/configuration/{siteId}/studio/translation/localeCodes` -> pass in the localeCodes JSON and it'll write it

For global configuration, it will be a reserved keyword, like `studio_root` or similar (since we already have that).

In the near-term, we will only support reading whole sections, and we can move to support writing and key access in the future. So only implement the GET.

This will change the UI behavior over time to think about config as params and not as files. This will help moves us towards our final goal of a new generalized configuration service independent of files.

src/main/webapp/repo-bootstrap/global/configuration/samples/sample-site-config.xml
sumerjabri: The sample won't be commented out.

src/main/resources/crafter/studio/upgrade/3.2.x/3.2.0/site/translation-config.xml
```diff
@@ -0,0 +1,55 @@
...
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~ GNU General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU General Public License
+  ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<translation-config>
+
+    <!-- List of codes for the supported locales -->
```
sumerjabri, line 20: Please don't use this type of commenting, it's too hard to fix without a high-end IDE.

src/main/webapp/repo-bootstrap/global/blueprints/1000_website_editorial/config/studio/translation-config.xml
```diff
@@ -0,0 +1,55 @@
...
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~ GNU General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU General Public License
+  ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<translation-config>
+
+    <!-- List of codes for the supported locales -->
```
sumerjabri, line 20: Please don't use this type of commenting, it's too hard to fix without a high-end IDE.

src/main/webapp/repo-bootstrap/global/blueprints/2000_headless_store/config/studio/translation-config.xml
```diff
@@ -0,0 +1,55 @@
...
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~ GNU General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU General Public License
+  ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<translation-config>
+
+    <!-- List of codes for the supported locales -->
```
sumerjabri, line 20: Please don't use this type of commenting, it's too hard to fix without a high-end IDE.

src/main/webapp/repo-bootstrap/global/blueprints/4000_empty/config/studio/translation-config.xml
```diff
@@ -0,0 +1,55 @@
...
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~ GNU General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU General Public License
+  ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<translation-config>
+
+    <!-- List of codes for the supported locales -->
```
sumerjabri, line 20: Please don't use this type of commenting, it's too hard to fix without a high-end IDE.

src/main/webapp/repo-bootstrap/global/blueprints/5000_headless_blog/config/studio/translation-config.xml
```diff
@@ -0,0 +1,55 @@
...
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~ GNU General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU General Public License
+  ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<translation-config>
+
+    <!-- List of codes for the supported locales -->
```
sumerjabri, line 20: Please don't use this type of commenting, it's too hard to fix without a high-end IDE.

sumerjabri (changes requested):
Please use 
style of commenting in XML to help people editing in Studio or not-so-fancy IDEs.

sumerjabri (approved):

src/main/resources/crafter/studio/studio-services-context.xml
```diff
@@ -280,6 +280,8 @@
     <bean id="studioEventService" class="org.craftercms.studio.impl.v1.service.event.EventServiceImpl" />
 
     <bean id="configurationService" class="org.craftercms.studio.impl.v2.service.configuration.ConfigurationServiceImpl">
+        <constructor-arg ref="crafter.configurationReader"/>
```
dejan-brkic, line 292: Why mixing constructor args and properties? It should be either all constructor agrs or all properties

dejan-brkic (changes requested):

src/main/resources/crafter/studio/studio-services-context.xml
jdrossl: I have been using constructor args for all new code because Spring is recommending it for required properties (they deprecated the `@Required` property that was used before) so in the future when we upgrade Spring it would be easier, I'm not sure if we want to refactor all existing code right now. @sumerjabri @avasquez614 any thoughts?

dejan-brkic (approved):

PR #2039, opened 2020-06-18
[4020] -Site Translation- Add locale as an intrinsic item metadata in the content modeling tool

jvega190:
https://github.com/craftercms/craftercms/issues/4020


rart (approved):

rart (approved):
Don't see any yaml updates. Not sure if they are needed for these.

PR #2066, opened 2020-07-06
Add ElasticSearch direct query

phuongnq:
Add an option to search with Elastic Search query:

Sample of query:

```
regexp:localId:/static-assets/.*,match:localId:lastEditedBy:admin
```

### Reason
Search API should have a method to directly query ElasticSearch.
Above I'm using `,` to separate each query and `:` to separate among operator, param and value.

If this solution does not fit, will it possible to discuss a similar solution?
In my case, I want to search with query such as:

```
{
    "query" : {
    	"bool": {
    		"must": [
    			{ 
    				"regexp": {
    					"localId": {
                			"value": "/static-assets/.*viber.*",
                			"flags" : "ALL",
                			"max_determinized_states": 10000,
                			"rewrite": "constant_score"
    					}
    				}
            	},
    			{ "match" : { "lastEditedBy": "admin" } }
    		]	
    	}
    }
}
```

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/service/search/internal/SearchServiceInternalImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/service/search/internal/SearchServiceInternalImpl.java
@@ -434,6 +434,21 @@ public class SearchServiceInternalImpl implements SearchServiceInternal {
             query.must(keywordsQuery);
         }
 
+        String esQueries = params.getEsQueries();
+        if (StringUtils.isNotEmpty(esQueries)) {
+            String[] keywords = esQueries.split(",");
+            if (ArrayUtils.isNotEmpty(keywords)) {
+                for (String keyword : keywords) {
+                    String[] queries = keyword.split(":");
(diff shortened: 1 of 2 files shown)


jdrossl:
hi @phuongnq the existing API already provides a field `query` that you can use to execute a custom query in Lucene's format https://app.swaggerhub.com/apis/craftercms/studio/3.1.8.2#/SearchParams

Here is an example:
```
{
  "query": "localId://static-assets/.*/ AND lastEditedBy:admin"
}
```

More info on the syntax of the query https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html

sumerjabri:
@phuongnq we discussed this internally, and we are unable to accept the PR since it introduces a custom syntax and opens things beyond what we want at this time. We will consider a PR that can open ES more, but remain consistent with Lucene/ES syntax and honors our security filters.

We hope the approach described by @jdrossl helps. If not, we offer a second option of using a WAR overlay to customize Studio to have your service.

We're also happy to discuss other ideas in Slack.

PR #2080, opened 2020-07-13
Refactor for text encryption updates

jdrossl:
https://github.com/craftercms/craftercms/issues/4178

Also fix some tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -920,6 +920,7 @@
                     </systemPropertyVariables>
                     <environmentVariables>
                         <CRAFTER_DATA_DIR>target/test-resources/data</CRAFTER_DATA_DIR>
+                        <MARIADB_ROOT_PASSWD>root</MARIADB_ROOT_PASSWD>
                     </environmentVariables>
                 </configuration>
             </plugin>
--- a/src/test/java/org/craftercms/studio/impl/v2/repository/blob/BlobAwareContentRepositoryTest.java
+++ b/src/test/java/org/craftercms/studio/impl/v2/repository/blob/BlobAwareContentRepositoryTest.java
@@ -25,7 +25,7 @@ import org.craftercms.studio.api.v1.to.DeploymentItemTO;
 import org.craftercms.studio.api.v1.to.VersionTO;
 import org.craftercms.studio.api.v2.repository.blob.StudioBlobStore;
 import org.craftercms.studio.api.v2.repository.blob.StudioBlobStoreResolver;
-import org.craftercms.studio.impl.v2.repository.GitContentRepository;
+import org.craftercms.studio.impl.v1.repository.git.GitContentRepository;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Captor;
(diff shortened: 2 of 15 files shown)


src/test/java/org/craftercms/studio/impl/v2/repository/blob/BlobAwareContentRepositoryTest.java
```diff
@@ -25,7 +25,7 @@
 import org.craftercms.studio.api.v1.to.VersionTO;
 import org.craftercms.studio.api.v2.repository.blob.StudioBlobStore;
 import org.craftercms.studio.api.v2.repository.blob.StudioBlobStoreResolver;
-import org.craftercms.studio.impl.v2.repository.GitContentRepository;
+import org.craftercms.studio.impl.v1.repository.git.GitContentRepository;
```
sumerjabri, line 28: Why?

src/test/resources/crafter/studio/extension/studio-config-override.yaml
```diff
@@ -3,4 +3,4 @@
 studio.db.port: 33307
 studio.db.socket: /tmp/MariaDB4j.33307.sock
 studio.db.url: jdbc:mariadb://127.0.0.1:33307/crafter?user=crafter&password=crafter
-studio.db.initializer.url: jdbc:mariadb://127.0.0.1:33307?user=root&password=
+studio.db.initializer.url: jdbc:mariadb://127.0.0.1:33307?user=root&password=${env:MARIADB_ROOT_PASSWD}
```
sumerjabri, line 6: Was this missing before and causing the tests to fail?

pom.xml
```diff
@@ -920,6 +920,7 @@
                     </systemPropertyVariables>
                     <environmentVariables>
                         <CRAFTER_DATA_DIR>target/test-resources/data</CRAFTER_DATA_DIR>
+                        <MARIADB_ROOT_PASSWD>root</MARIADB_ROOT_PASSWD>
```
sumerjabri, line 923: Was this missing before and causing problems?
Also, doesn't Studio now automatically change the password if it's root?

sumerjabri (changes requested):

src/test/java/org/craftercms/studio/impl/v2/repository/blob/BlobAwareContentRepositoryTest.java
jdrossl: This was caused by a bad merge and the test was not compiling

src/test/resources/crafter/studio/extension/studio-config-override.yaml
jdrossl: yeah

pom.xml
jdrossl: The default configuration is to use the env variable so for the test it has to be provided in the config, I don't know if it is changed later but it was preventing the test from running

sumerjabri (approved):

PR #2084, opened 2020-07-17
[4189] [studio] Publisher doesn't work when a site with a published repo is removed and then a new site is created with same site name.

dejan-brkic:
Fixed issue caused by side-effect of lazy published repo

https://github.com/craftercms/craftercms/issues/4189


src/test/java/org/craftercms/studio/impl/v2/repository/blob/BlobAwareContentRepositoryTest.java
```diff
@@ -25,7 +25,7 @@
 import org.craftercms.studio.api.v1.to.VersionTO;
 import org.craftercms.studio.api.v2.repository.blob.StudioBlobStore;
 import org.craftercms.studio.api.v2.repository.blob.StudioBlobStoreResolver;
-import org.craftercms.studio.impl.v1.repository.git.GitContentRepository;
+import org.craftercms.studio.impl.v2.repository.GitContentRepository;
```
sumerjabri, line 28: Why? This breaks the system.

sumerjabri (changes requested):

src/test/java/org/craftercms/studio/impl/v2/repository/blob/BlobAwareContentRepositoryTest.java
dejan-brkic: Because below in the code V1 is declared with the full package name and V2 just class name (package is imported). Somewhere in history was a bad merge. Lines 75-78

sumerjabri: Did you fix the unit test to work with v2?

dejan-brkic: yes

sumerjabri (approved):

PR #2096, opened 2020-07-23
Fix login redirect with proxies or load balancers

jdrossl:
craftercms/craftercms#4205

Depends on https://github.com/craftercms/commons/pull/237

### DIFF
--- a/src/main/webapp/WEB-INF/web.xml
+++ b/src/main/webapp/WEB-INF/web.xml
@@ -67,6 +67,16 @@
     </filter>
 
     <filter>
+        <filter-name>cloudfrontForwardedHeaderFilter</filter-name>
+        <filter-class>org.springframework.web.filter.DelegatingFilterProxy</filter-class>
+    </filter>
+
+    <filter>
+        <filter-name>forwardedHeaderFilter</filter-name>
+        <filter-class>org.springframework.web.filter.DelegatingFilterProxy</filter-class>
+    </filter>
+
+    <filter>
         <filter-name>springSecurityFilterChain</filter-name>
         <filter-class>org.springframework.web.filter.DelegatingFilterProxy</filter-class>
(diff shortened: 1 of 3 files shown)


src/main/webapp/WEB-INF/web.xml
```diff
@@ -115,6 +125,16 @@
         <url-pattern>/*</url-pattern>
     </filter-mapping>
 
+    <filter-mapping>
+        <filter-name>cloudfrontForwardedHeaderFilter</filter-name>
+        <url-pattern>/*</url-pattern>
+    </filter-mapping>
+
+    <filter-mapping>
```
avasquez614, line 133: Is it possible to make these filters optional? That can only be turned on with a flag in the Studio configuration? Because it could be an issue, if the load balancer or reverse proxy doesn't overwrite the flags, or a client is using no reverse proxy.

avasquez614 (approved):

src/main/webapp/WEB-INF/web.xml
```diff
@@ -115,6 +125,16 @@
         <url-pattern>/*</url-pattern>
     </filter-mapping>
 
+    <filter-mapping>
```
avasquez614, line 128: Is it possible to make these filters optional? That can only be turned on with a flag in the Studio configuration? Because it could be an issue, if the load balancer or reverse proxy doesn't overwrite the flags, or a client is using no reverse proxy.

avasquez614 (changes requested):

src/main/webapp/WEB-INF/web.xml
jdrossl: Both filters check the request and are skipped if the headers are not present, do you think we sill need a flag to disable them?

avasquez614 (approved):

PR #2101, opened 2020-07-28
Fixed UM issues

dejan-brkic:
Fixed UM issues

https://github.com/craftercms/craftercms/issues/4213
https://github.com/craftercms/craftercms/issues/4214

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
@@ -191,6 +191,7 @@ public class SiteServiceImpl implements SiteService {
     protected SitesServiceInternal sitesServiceInternal;
     protected AuditServiceInternal auditServiceInternal;
     protected ConfigurationService configurationService;
+    @Autowired
     protected SiteFeedMapper siteFeedMapper;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
```diff
@@ -191,6 +191,7 @@
     protected SitesServiceInternal sitesServiceInternal;
     protected AuditServiceInternal auditServiceInternal;
     protected ConfigurationService configurationService;
+    @Autowired
```
jdrossl, line 194: I think we don't have `@Autowired` anywhere else in Studio right? In theory the set should work in the same way, do you know if this is a specific issue with MyBatis?

dejan-brkic: I removed `@Autowired` 2 weeks ago to align it with other mappers. Apparently that introduced NPE because `siteFeedMapper` is not set for UM execution. I was not able to provide proper order of bean construction with the `depends-on` attribute. `@Autowired` has to remain until we find a proper solution for bean creation order

jdrossl (approved):

PR #2106, opened 2020-07-29
Upgrade Manager:

jdrossl:
Re-encrypt:
- Properties in configuration files
- Credentials in remote repositories
- Credentials in cluster members

https://github.com/craftersoftware/craftercms/issues/145

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/db/DbEncryptionUpgradeOperation.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/db/DbEncryptionUpgradeOperation.java
@@ -1,0 +1,143 @@
+/*
+ * Copyright (C) 2007-2020 Crafter Software Corporation. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as published by
+ * the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.craftercms.studio.impl.v2.upgrade.operations.db;
+
+import org.craftercms.commons.crypto.CryptoException;
+import org.craftercms.commons.crypto.TextEncryptor;
+import org.craftercms.studio.api.v1.log.Logger;
+import org.craftercms.studio.api.v1.log.LoggerFactory;
+import org.craftercms.studio.api.v2.dal.ClusterMember;
+import org.craftercms.studio.api.v2.dal.RemoteRepository;
+import org.craftercms.studio.api.v2.exception.UpgradeException;
+import org.craftercms.studio.impl.v2.upgrade.operations.AbstractUpgradeOperation;
+import org.springframework.jdbc.core.BeanPropertyRowMapper;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+import java.util.List;
+
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Implementation of {@link org.craftercms.studio.api.v2.upgrade.UpgradeOperation} that upgrades encrypted values in
+ * the database.
+ *
+ * @author joseross
+ * @since 3.1.9
+ */
+public class DbEncryptionUpgradeOperation extends AbstractUpgradeOperation {
+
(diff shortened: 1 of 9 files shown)


src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/db/DbEncryptionUpgradeOperation.java
```diff
@@ -0,0 +1,143 @@
...
+
+    protected String REMOTE_REPOSITORIES_QUERY = "select id, remote_password, remote_token, remote_private_key, " +
+            "authentication_type from remote_repository where authentication_type != ?";
+    protected String REMOTE_REPOSITORIES_UPDATE = "update remote_repository set remote_password = ?, " +
+            "remote_token = ?, remote_private_key = ? where id = ?";
+
+    protected String CLUSTER_MEMBERS_QUERY = "select id, git_password, git_token, git_private_key, git_auth_type " +
+            "from cluster where git_auth_type != ?";
+    protected String CLUSTER_MEMBERS_UPDATE = "update cluster set git_password = ?, git_token = ?, " +
+            "git_private_key = ? where id = ?";
```
dejan-brkic, line 52: Why using a parametrized select query if we are going to use always the same value for the parameter?
Why use `?` for parameters? It is prone to SQL injection and with multiple parameters, it is prone to an error caused by the wrong parameter order.

dejan-brkic (changes requested):

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/db/DbEncryptionUpgradeOperation.java
jdrossl: For the `authentication_type` since we already have a constant in the code I used the parameter to avoid hardcoding the value.

About the use of `?` for parameters I don't see any issues since this operation will only transform existing values from the DB (if there is a SQL injection it's because the DB was already compromised) and this is a one time operation so neither the SQL or the code will be changing later (that's why I choose to use the simple parameters instead of named parameters or the DAO classes)

dejan-brkic: Having a full query as constant is a more readable code, even if you use existing constant by concatenating string. Besides, there is a mater of Jdbc having to parse and replace the parameter value.

Somebody can extend your class and create malicious code that can exploit SQL injection. It is not only about how are you using an operation. Additionally, we should write a cleaner and safer code even where it is not necessary to create good habits and good practice.

src/main/resources/crafter/studio/upgrade/pipelines.yaml
```diff
@@ -1029,6 +1034,14 @@ pipelines:
           pattern: previewUrl
           replacement: authoringUrl
         - type: versionFileUpgrader
+    - currentVersion: 3.1.7.3
+      nextVersion: 3.1.9.0
```
dejan-brkic: 3.1.9.0 vs 3.1.9.1 and 3.1.9.2 on lines 835 and 836? Two different pipelines yes, but shouldn't they at least be in sync for next version value when it changes? If I remember well it can be only 3.1.9 and not 3.1.9.0 by naming convention.

dejan-brkic (changes requested):

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/db/DbEncryptionUpgradeOperation.java
jdrossl: I disagree because string concatenations are less secure than `?` parameters, also someone extending classes will have full access to the DB a single SQL statement will not make any difference in that case

jdrossl: @sumerjabri @avasquez614 maybe we can have a call do review these concerns

dejan-brkic (approved):

src/main/java/org/craftercms/studio/impl/v2/upgrade/operations/db/DbEncryptionUpgradeOperation.java
sumerjabri: The `?` is fine since it's not SQL injectable. We're passing in the constant that we defined for this purpose. I don't see a class extension as an attack vector.

src/main/resources/crafter/studio/upgrade/pipelines.yaml
sumerjabri: Let's continue with our convention of 3.1.9 and then 3.1.9.1 for the following. This means the last .0 is assumed since we don't always have it. Perhaps this can be revisited in the future to always have .0 at the beginning. I don't have strong feelings about this.

Regarding keeping the pipelines in sync, I don't see that as required nor necessary, it's busy work with no real value.

sumerjabri (approved):

PR #2191, opened 2020-09-24
[4299] [studio] Cluster: When a site created from remote, the remote is only added in the node that created the site

dejan-brkic:
Remote repository is propagated to all cluster nodes

https://github.com/craftercms/craftercms/issues/4299


sumerjabri (changes requested):
Please use the DB retry mechanism.

sumerjabri (approved):

PR #2194, opened 2020-09-25
[4315] [studio] Update the publisher to handle null commit_ids

dejan-brkic:
Fixed issue with publishing null commit id

https://github.com/craftercms/craftercms/issues/4315

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/deployment/PublishingManagerImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/deployment/PublishingManagerImpl.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import org.apache.commons.lang3.StringUtils;
@@ -215,7 +216,7 @@ public class PublishingManagerImpl implements PublishingManager {
                     LOGGER.debug("Environment is live, transition item to LIVE state {0}:{1}", site, path);
 
                     // check if commit id from workflow and from object state match
-                    if (itemMetadata.getCommitId() != null && itemMetadata.getCommitId().equals(item.getCommitId())) {
+                    if (Objects.isNull(itemMetadata.getCommitId()) || itemMetadata.getCommitId().equals(item.getCommitId())) {
                         objectStateService.transition(site, contentItem, TransitionEvent.DEPLOYMENT);
                     }
                 } else {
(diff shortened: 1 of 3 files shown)


src/main/java/org/craftercms/studio/impl/v2/repository/GitContentRepository.java
```diff
@@ -964,8 +964,12 @@ public void publish(String site, String sandboxBranch, List<DeploymentItemTO> de
                         commitId = deploymentItem.getCommitId();
                         path = helper.getGitPath(deploymentItem.getPath());
                         if (Objects.isNull(commitId)) {
-                            logger.warn("Skipping file " + path + " because commit id is null");
-                            continue;
+                            if (contentExists(site, path)) {
+                                commitId = getRepoLastCommitId(site);
```
sumerjabri, line 968: Will this guarantee the file being present in the index and a successful publish?

src/main/java/org/craftercms/studio/impl/v2/repository/GitContentRepository.java
```diff
@@ -964,8 +964,12 @@ public void publish(String site, String sandboxBranch, List<DeploymentItemTO> de
...
+                            } else {
+                                logger.warn("Skipping file " + path + " because commit id is null");
```
sumerjabri, line 970: If the file doesn't exist on disk, it means it exists only in the database. Should this trigger removal from the database (including workflow/etc.)?

src/main/java/org/craftercms/studio/impl/v1/service/deployment/PublishingManagerImpl.java
```diff
@@ -215,7 +216,7 @@ else if (StringUtils.equalsIgnoreCase(LIVE_ENVIRONMENT, item.getEnvironment()) |
                     LOGGER.debug("Environment is live, transition item to LIVE state {0}:{1}", site, path);
 
                     // check if commit id from workflow and from object state match
-                    if (itemMetadata.getCommitId() != null && itemMetadata.getCommitId().equals(item.getCommitId())) {
+                    if (Objects.isNull(itemMetadata.getCommitId()) || itemMetadata.getCommitId().equals(item.getCommitId())) {
```
sumerjabri, line 219: What does `itemMetadata.getCommitId().equals(item.getCommitId())` mean?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v2/repository/GitContentRepository.java
dejan-brkic: yes

src/main/java/org/craftercms/studio/impl/v2/repository/GitContentRepository.java
dejan-brkic: we agreed that it should be skipped. no other action

src/main/java/org/craftercms/studio/impl/v1/service/deployment/PublishingManagerImpl.java
dejan-brkic: that is the foundation for multiple publish requests of the same file (multiple versions, it will have status "LIVE" only if the version being published is the one in item metadata (the latest version)

sumerjabri (approved):

PR #2195, opened 2020-09-25
[4315] [studio] Update the publisher to handle null commit_ids

dejan-brkic:
Fixed issue with publishing null commit id

https://github.com/craftercms/craftercms/issues/4315


sumerjabri (changes requested):
See comments in `support/3.1.x`

sumerjabri (approved):

PR #2196, opened 2020-09-25
Fix for duplicated descriptors:

jdrossl:
- Change filename modifier to `-copy-N`
- Change internal-name modifier to `(Copy N)`
- Replace existing modifiers instead of appending new ones
- Keep track of existing copies to properly increment the modifier

https://github.com/craftercms/craftercms/issues/4061

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
@@ -182,6 +182,12 @@ public class ContentServiceImpl implements ContentService {
     public final static Pattern COPY_FILE_PATTERN = Pattern.compile("(.+)-([0-9]+)\\.(.+)");
     public final static Pattern COPY_FOLDER_PATTERN = Pattern.compile("(.+)-([0-9]+)");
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
```diff
@@ -182,6 +182,12 @@
     public final static Pattern COPY_FILE_PATTERN = Pattern.compile("(.+)-([0-9]+)\\.(.+)");
     public final static Pattern COPY_FOLDER_PATTERN = Pattern.compile("(.+)-([0-9]+)");
 
+    public final static Pattern COPY_FILE_MODIFIER_PATTERN = Pattern.compile(".+(-copy-([\\d]+))(.+)?(\\..*)?");
```
sumerjabri, line 185: What if it's a paste operation/upload/bulk-upload/create and not a duplicate? That operation should result in the name "${old-name}\ (n)" and not have the word `copy` in it.

sumerjabri (changes requested):

sumerjabri (approved):

PR #2200, opened 2020-09-29
Validate filename & path for content operations

jdrossl:
https://github.com/craftercms/craftercms/issues/4060

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
@@ -88,7 +88,6 @@ import org.craftercms.studio.api.v2.dal.Item;
 import org.craftercms.studio.api.v2.dal.ItemState;
 import org.craftercms.studio.api.v2.exception.validation.FilenameTooLongException;
 import org.craftercms.studio.api.v2.exception.validation.PathTooLongException;
-import org.craftercms.studio.api.v2.exception.validation.ValidationException;
 import org.craftercms.studio.api.v2.service.audit.internal.AuditServiceInternal;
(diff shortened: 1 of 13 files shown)


src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
```diff
@@ -2586,6 +2603,18 @@ public boolean pullFromRemote(String siteId, String remoteName, String remoteBra
         return toRet;
     }
 
+    //TODO: Move this to a fixed always-on site policy when implemented
+    protected void validateContent(String path, String filename) throws ServiceLayerException {
+        if (filename.length() >= 255) {
```
sumerjabri: Please put the numbers 255, 1024 as configuration items in studio-config.yaml. This will help us configure the DB and Studio more easily if we have to, and avoids the magic number here. I expect this configuration to be in RAM and not pulled from the configuration file everytime.

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
jdrossl: We can make it configurable for the service, but not for DB because those are in the create schema script and it's not possible to inject them there

sumerjabri: Yes, understood. Go ahead and make configurable, please. Also, make a note in the configuration file that DB changes are also required. Please don't propagate this to studio-config-override, only keep studio-config.yaml.

sumerjabri (approved):

PR #2208, opened 2020-10-06
Fixed response formatting and data issues

dejan-brkic:
Fixed response formatting and data issues


sumerjabri (changes requested):
* make this a variable: parentItem.getUri() + FILE_SEPARATOR + fileName
* add a helper that does the work for the creation of item and upsert and centralize this functionality

sumerjabri (approved):

PR #2212, opened 2020-10-12
[4335] [studio] Renaming a file to the same name with a different case corrupts the database

dejan-brkic:
Fixed rename with case only change

https://github.com/craftercms/craftercms/issues/4335

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
@@ -1028,7 +1028,9 @@ public class ContentServiceImpl implements ContentService {
 
         if (!renamedItem.isFolder()) {
             if (objectMetadataManager.metadataExist(site, movePath)) {
-                objectMetadataManager.deleteObjectMetadata(site, movePath);
+                if (!StringUtils.equalsIgnoreCase(fromPath, movePath)) {
+                    objectMetadataManager.deleteObjectMetadata(site, movePath);
+                }
             }
             objectMetadataManager.updateObjectPath(site, fromPath, movePath);
         }
--- a/src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
@@ -1163,7 +1163,7 @@ public class SiteServiceImpl implements SiteService {
                         } else {
                             logger.debug("Set item state for site: " + site + " path: " + repoOperation.getPath());
                             // If state already exists (renamed item with only case change) and path in the state
-                            // table is equal except for the case, than update item state table with new path value
+                            // table is equal except for the case, then update item state table with new path value
                             // (DB by default is case insensitive)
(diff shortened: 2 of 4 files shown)


src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
```diff
@@ -1028,7 +1028,9 @@ protected void updateDatabaseOnMove(String site, String fromPath, String movePat
 
         if (!renamedItem.isFolder()) {
             if (objectMetadataManager.metadataExist(site, movePath)) {
-                objectMetadataManager.deleteObjectMetadata(site, movePath);
+                if (!StringUtils.equalsIgnoreCase(fromPath, movePath)) {
+                    objectMetadataManager.deleteObjectMetadata(site, movePath);
```
sumerjabri, line 1032: Should this be using the new Item mechanics?

src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
```diff
@@ -1166,12 +1162,18 @@ public boolean syncDatabaseWithRepo(@ValidateStringParam(name = "site") String s
                             objectStateService.insertNewEntry(site, repoOperation.getPath());
                         } else {
                             logger.debug("Set item state for site: " + site + " path: " + repoOperation.getPath());
+                            if (StringUtils.equalsIgnoreCase(state.getPath(), repoOperation.getPath()) &&
+                                    !StringUtils.equals(state.getPath(), repoOperation.getPath())) {
```
sumerjabri, line 1169: Add a comment to explain the logic here. My read is: the source is equal to the destination if and only if case is ignored.

src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
```diff
@@ -1166,12 +1162,18 @@ public boolean syncDatabaseWithRepo(@ValidateStringParam(name = "site") String s
...
+                                objectStateService.updateObjectPath(site, state.getPath(), repoOperation.getPath());
+                            }
                             objectStateService.transition(site, repoOperation.getPath(), TransitionEvent.SAVE);
                         }
 
                         logger.debug("Set item metadata for site: " + site + " path: " + repoOperation.getPath());
                         if (!objectMetadataManager.metadataExist(site, repoOperation.getPath())) {
                             objectMetadataManager.insertNewObjectMetadata(site, repoOperation.getPath());
+                        } else {
+                            objectMetadataManager.updateObjectPath(site, repoOperation.getPath(), repoOperation.getPath());
```
sumerjabri, line 1179: Should we using the new tables?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
sumerjabri: Another way to say it:

// If they're equal with mixed case, then update ...

sumerjabri (approved):

PR #2224, opened 2020-10-22
[3956] [studio] Parse and cache `permission-mappings-config.xml`

dejan-brkic:
Added caching for available actions

https://github.com/craftercms/craftercms/issues/3956


sumerjabri (changes requested):
Please refactor as we discussed.

sumerjabri (changes requested):
Please update per our conversation:
* Switch to something like this:
public static final long ALL_PERMISSION =                       1111111111111111111111111111111111111111111111111111111111111111B;
* Remove the enums in favor of this pattern
* Weave in the target in the annotation

src/main/java/org/craftercms/studio/api/v2/dal/security/SitePermissionMappings.java
```diff
@@ -48,6 +52,7 @@ public long getAvailableActions(String username, List<Group> groups, String path
 
         long availableActions = 0L;
         for (String role : rolesList) {
+            logger.error("Role = " + role);
```
sumerjabri: What's this for?

sumerjabri (approved):

PR #2241, opened 2020-11-05
Initial Site Policy Implementation:

jdrossl:
- Add sample & config file
- Integrate with content service

https://github.com/craftercms/craftercms/issues/4354

### DIFF
--- a/src/main/java/org/craftercms/studio/model/policy/Action.java
+++ b/src/main/java/org/craftercms/studio/model/policy/Action.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright (C) 2007-2020 Crafter Software Corporation. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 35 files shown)


src/main/java/org/craftercms/studio/model/policy/Action.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * The content-type of the action
+     */
+    protected String contentType;
+
+    /**
+     * The file size of the action
+     */
+    protected long fileSize;
```
avasquez614: I think metadata like this property (and content type) might be better to have in a map called `contentMetadata`. In the future, we might need other content properties like mime type, and I think it doesn't make sense to add a bunch of properties here when they are not used all the time and/or depend on the type of content (item vs static asset).

avasquez614 (changes requested):

sumerjabri (approved):

PR #2286, opened 2020-11-23
Add new API for clipboard operations

jdrossl:
https://github.com/craftercms/craftercms/issues/4379

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -60,8 +60,6 @@ tags:
     description: Translation operations
   - name: workflow
     description: Workflow operations
-  - name: clipboard
-    description: Clipboard operations
 
 paths:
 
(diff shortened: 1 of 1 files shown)


src/main/api/2/studio-api-2.yaml
```diff
@@ -4592,6 +4594,131 @@ paths:
...
+          application/json:
+            schema:
+              type: object
+              properties:
+                siteId:
+                  type: string
+                  description: The id of the site
+                path:
+                  type: string
+                  description: The path of the item to duplicate
```
rart: Will duplicate always only duplicate the individual item (i.e. it won't duplicate children ever)?

src/main/api/2/studio-api-2.yaml
```diff
@@ -6168,6 +6295,36 @@ components:
...
+        path:
+          type: string
+          description: The path of the item
+        children:
+          type: array
+          description: The child items
+          items:
+            $ref: '#/components/schemas/ClipboardItem'
+      required:
+        - path
```
rart: Are `children` meant to be the child items that were selected to be copied too?

This hierarchical structure makes it a bit complex to build the request, plus the request body will be larger. Would simply having a list of paths not be sufficient? Paths bear hierarchy on their own.

src/main/api/2/studio-api-2.yaml
```diff
@@ -6168,6 +6295,36 @@ components:
...
+      type: object
+      properties:
+        siteId:
+          type: string
+          description: The id of the site
+        targetPath:
+          type: string
+          description: The target path for the operation
+        item:
+          $ref: '#/components/schemas/ClipboardItem'
```
rart: See previous comment about hierarchy. Could we stick with a list of paths (strings)?

src/main/api/2/studio-api-2.yaml
```diff
@@ -4592,6 +4594,131 @@ paths:
...
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+                  items:
+                    description: The list of the new items
```
rart: Is the copy operation actually the paste from a prior copy? Same goes for cut spec below. Otherwise, how would we the have the list of new items already if we haven't pasted?

If this is the case, I think even though at a disk/git level copy and cut are the actual ops (i.e. cp, mv), in this context ("clipboard") the mental model is more copy + paste. Isn't `paste` named copy/cut unexpected? Shall we consider having a single `/clipboard/paste` with a `request.type = 'copy' | 'cut'`. Or if we wish to keep them separate, perhaps `/clipboard/paste_from_copy` and  `/clipboard/paste_from_cut`?

If we are going to keep the clipboard on the UI and the BE only performs the mv/cp, perhaps the end points belong rather to `/content` and there, `copy` and `move` would be clearer. `duplicate` also fits well in `/content`.

src/main/api/2/studio-api-2.yaml
jdrossl: @sumerjabri ^

src/main/api/2/studio-api-2.yaml
jdrossl: This API was modeled following the existing one, there is a lot of logic on the backend that already expects this structure

src/main/api/2/studio-api-2.yaml
jdrossl: @rart yes, the UI will need to hold the items and know if the operation to perform is a copy or cut, a single call will be made to the backend

rart (approved):

src/main/api/2/studio-api-2.yaml
```diff
@@ -3486,6 +3486,92 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /content/paste:
+    post:
+      tags:
+        - content
+      summary: Copy or cut the given items to the target path
```
sumerjabri: Paste a tree of items to a target path (copy content or move content). To be used by the UI in the clipboard implementation.

src/main/api/2/studio-api-2.yaml
```diff
@@ -3486,6 +3486,92 @@ paths:
...
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+                  items:
+                    description: The list of the new items
```
sumerjabri: The list of pasted items

sumerjabri (changes requested):

sumerjabri (approved):

PR #2294, opened 2020-11-27
Implement new Clipboard API v2

jdrossl:
https://github.com/craftercms/craftercms/issues/4379


src/main/java/org/craftercms/studio/impl/v2/service/clipboard/internal/ClipboardServiceInternalImpl.java
```diff
@@ -0,0 +1,109 @@
...
+                        // recurse on copied children
+                        if (isNotEmpty(item.getChildren())) {
+                            pasteItemsInternal(siteId, operation, newPath, item.getChildren(), pastedItems);
+                        }
+                        break;
+                    default:
+                        logger.warn("Unsupported clipboard operation {0}", operation);
+                }
+
+                pastedItems.add(newPath);
```
sumerjabri, line 90: This means you only return the root item, correct?

src/main/java/org/craftercms/studio/impl/v2/service/clipboard/internal/ClipboardServiceInternalImpl.java
```diff
@@ -0,0 +1,109 @@
...
+                        newPath = contentService.moveContent(siteId, item.getPath(), targetPath);
+                        break;
+                    case COPY:
+                        // RDTMP_COPYPASTE
+                        // CopyContent interface is able to send status and new path yet
+                        newPath = contentService.copyContent(siteId, item.getPath(), targetPath);
+
+                        // recurse on copied children
+                        if (isNotEmpty(item.getChildren())) {
+                            pasteItemsInternal(siteId, operation, newPath, item.getChildren(), pastedItems);
```
sumerjabri, line 83: Nothing is returned, therefore the children's paths are not being returned.

src/main/java/org/craftercms/studio/impl/v2/service/clipboard/internal/ClipboardServiceInternalImpl.java
jdrossl: A list is always returned even if only one item was copied/moved

src/main/java/org/craftercms/studio/impl/v2/service/clipboard/internal/ClipboardServiceInternalImpl.java
jdrossl: The children's paths are added to the list (last parameter of the method)

src/main/java/org/craftercms/studio/impl/v2/service/clipboard/internal/ClipboardServiceInternalImpl.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+
+    public String duplicateItem(String siteId, String path) throws ServiceLayerException, UserNotFoundException {
+        // Get the parent url: for folders & components it's just parent, for pages it's the parent of the parent
+        var filename = getName(path);
+        // TODO: Find a better way to do this?
+        var parentUrl = "index.xml".equals(filename)? getFullPathNoEndSeparator(getFullPathNoEndSeparator(path))
+                                                            : getFullPathNoEndSeparator(path);
+        var item = contentService.getContentItem(siteId, parentUrl);
```
sumerjabri: Set depth and make it limited to reduce cost.

sumerjabri (changes requested):
Set depth (see notes).

sumerjabri (approved):

PR #2295, opened 2020-11-27
Add new API for User Properties

jdrossl:
https://github.com/craftercms/craftercms/issues/4423

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '3.2.0.5'
+  version: '3.2.0.6'
   title: Crafter Studio
   termsOfService: ''
   contact:
@@ -1126,8 +1126,10 @@ paths:
                   properties:
                     type: object
                     additionalProperties:
-                      type: string
-                    description: The properties of the current user
+                      type: object
+                      additionalProperties:
+                        type: string
+                    description: The properties of the current user grouped by site
         '400':
           $ref: '#/components/responses/BadRequest'
         '401':
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -1100,6 +1100,89 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /users/me/properties:
+    get:
+      tags:
+        - users
+      summary: Retrieve properties for the current user
+      description: 'Required Permission: "LOGGED_IN"'
+      operationId: getUserProperties
+      parameters:
+        - name: siteId
+          schema:
+            type: string
+          in: query
+          description: The id of the site (defaults to global)
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+                  properties:
+                    type: object
+                    additionalProperties:
+                      type: string
+                    description: The properties of the current user
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '500':
+          $ref: '#/components/responses/InternalServerError'
+    post:
+      tags:
+        - users
+      summary: Update properties for the current user
+      description: 'Required Permission: "LOGGED_IN"'
+      operationId: updateUserProperties
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                siteId:
+                  type: string
+                  description: The id of the site (defaults to global)
+                upsert:
+                  type: object
+                  additionalProperties:
+                    type : string
+                  description: Properties to update or add
+                delete:
+                  type: array
+                  items:
+                    type: string
+                  description: List of keys to delete
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+                  properties:
+                    type: object
+                    additionalProperties:
+                      type: string
+                    description: |-
+                      The properties of the current user after the updates
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '500':
+          $ref: '#/components/responses/InternalServerError'
+
   /users/forgot_password:
     get:
       tags:


src/main/api/2/studio-api-2.yaml
```diff
@@ -1100,6 +1100,89 @@ paths:
...
+        - users
+      summary: Retrieve properties for the current user
+      description: 'Required Permission: "LOGGED_IN"'
+      operationId: getUserProperties
+      parameters:
+        - name: siteId
+          schema:
+            type: string
+          in: query
+          description: The id of the site (defaults to global)
```
rart, line 1115: - Can we have an optional param to get only matching `property_key`s?
- It would be powerful and maximally flexible if we make `siteId` optional too; it would allow retrieving the same property of all sites or other things that could be valuable for an future settings screen or a property management plugin. If nothing is supplied the service could return all props for _this_ user.
- Should this be a paginated API?

src/main/api/2/studio-api-2.yaml
```diff
@@ -1100,6 +1100,89 @@ paths:
...
+              schema:
+                type: object
+                properties:
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+                  properties:
+                    type: object
+                    additionalProperties:
+                      type: string
+                    description: The properties of the current user
```
rart: 1. Isn't this an array of objects? (e.g. array of UserProperty schemas?)

2. What is the property object like? Based of the db schema, it'd be
```ts
interface UserProperty {
  id: number;
  key: string;
  value: string;
  siteId: string;
}
```

src/main/api/2/studio-api-2.yaml
```diff
@@ -1100,6 +1100,89 @@ paths:
...
+                upsert:
+                  type: object
+                  additionalProperties:
+                    type : string
+                  description: Properties to update or add
+                delete:
+                  type: array
+                  items:
+                    type: string
+                  description: List of keys to delete
```
rart, line 1164: Perhaps here we could have upsert/delete both be arrays of _UserProperty_ objects and `siteId` be sent as part of the objects so we could CRUD properties across sites in a single request?

src/main/api/2/studio-api-2.yaml
```diff
@@ -1100,6 +1100,89 @@ paths:
...
+              type: object
+              properties:
+                siteId:
+                  type: string
+                  description: The id of the site (defaults to global)
+                upsert:
+                  type: object
+                  additionalProperties:
+                    type : string
+                  description: Properties to update or add
```
rart, line 1157: Isn't this an array of UserProperty objects?

src/main/api/2/studio-api-2.yaml
sumerjabri: +1

src/main/api/2/studio-api-2.yaml
sumerjabri: @jdrossl this needs to allow the query of a single "key" within a site or across sites.

The DB is indexed on the "key" as well as "site".

I don't think this API should paginate.

src/main/api/2/studio-api-2.yaml
sumerjabri: It should be similar to what @rart shows, and similar to the DB.

sumerjabri (changes requested):
See notes.

src/main/api/2/studio-api-2.yaml
jdrossl: @rart `siteId` is optional, that is why it says that defaults to global

@sumerjabri all existing APIs use an empty `siteId` to reference the global repo, I'm not sure if changing that to mean all sites is a good idea. Maybe we can change it to be a keyword like `siteId: 'ANY'`

src/main/api/2/studio-api-2.yaml
jdrossl: I don't think this API should expose the full db table, I was thinking a very simple key-property GET/UPDATE/DELETE

src/main/api/2/studio-api-2.yaml
jdrossl: As I mentioned here https://github.com/craftercms/studio/pull/2295#discussion_r533432653 this would be the first cross-site operation we implement so I think we need to discuss the design further to make sure it doesn't have conflict with the existing APIs

src/main/api/2/studio-api-2.yaml
sumerjabri: @rart what's an example where you need to scan for a property that's across sites? (global properties will flow to the global (system) repo)

rart: A concrete example: a global preference that has a site override. We would get all and apply/prevail the current site if there are overrides.

Overall, I’m thinking around the flexibility of the API for different uses. Don’t know how open or closed we want it.

src/main/api/2/studio-api-2.yaml
rart: i don’t mind not getting the db id but feels like at the min we need the key, value, site triad.

src/main/api/2/studio-api-2.yaml
rart: It feels clear and non-conflicting given that the site id is moved into each object. Typically siteId is a top level, single param.

src/main/api/2/studio-api-2.yaml
sumerjabri: What's an example of such a property? (When we open things up we incur a cost to implement the general, I need to see the use-case more concretely.)

sumerjabri: @rart ^

jdrossl: @rart how are you planning handling global vs site specific? will the UI set global defaults for all users and then let them set site specific values? how will users go back to the global defaults? how would a user change the global default keeping the site specifics?

I had in mind a very simple key-pair API that is always site aware, as that is the usual schema of our API

rart: @sumerjabri an example: a prop that allows customising the widgets of my sidebar just for myself and wanting a particular override for one or more sites.

(Today, widgets are an xml file that applies to everyone but with user properties we could allow more personalised configs.) Same principle could go for Dashboards, etc (e.g. I want to add/remove a widget for my view of dashboard for all sites or for a given site.)

This may not be 3.2. Am I thinking too big?

I think for 3.2, mostly we’ll use this to move the stuff we keep in local browser storage to user props so they persist across terminals and browser data clearing. These are things like: language, UI theme(s), state of pages widgets (expanded/collapsed, folder I’m at, etc), current sidebar page I’m on, preview compatibility determination, etc.

@jdrossl those are product/UX questions that don’t all have an answer just yet. We need some internal product discussions to finalise but my intention is to enable these sorts of things and empower the plugin world.

sumerjabri: We will do a `*` for all sites down the road when needed. For now, not sending `siteId` (or "" or NULL) means global, otherwise, spec out the site.

src/main/api/2/studio-api-2.yaml
sumerjabri: response: {
  site1: {
    key: value
    key: value
  }
  site2: {
    key: value
    key: value
  }
}

sumerjabri (approved):

PR #2321, opened 2020-12-23
[206E] [studio] Optimize Commit Processing

dejan-brkic:
Commit processing optimization

https://github.com/craftersoftware/craftercms/issues/206


src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
```diff
@@ -744,19 +782,40 @@ private void createSiteCloneRemote(String siteId, String sandboxBranch, String d
 
                 upgradeManager.upgradeSite(siteId);
 
-                insertCreateSiteAuditLog(siteId, siteId);
-
                 // Add default groups
                 logger.info("Adding default groups for site " + siteId);
                 addDefaultGroupsForNewSite(siteId);
 
+                String lastCommitId = contentRepositoryV2.getRepoLastCommitId(siteId);
```
sumerjabri: Change to not create an audit log. Instead, set the markers such that it will calculate from the beginning.

sumerjabri (approved):

PR #2377, opened 2021-02-02
[4498] [studio] Add bulk getSandboxItems API calls

dejan-brkic:
Implemented getSandboxItemsBy* APIs

https://github.com/craftercms/craftercms/issues/4498


src/main/api/2/studio-api-2.yaml
```diff
@@ -3581,6 +3581,110 @@ paths:
...
+          in: query
+          description: item ids to get
+          required: true
+          schema:
+            type: array
+            items:
+              type: string
+        - name: preferContent
+          in: query
+          description: return content item in case if it exsits when quering for container
```
sumerjabri: when set to true, return an item instead of a folder if the path can match either

src/main/api/2/studio-api-2.yaml
```diff
@@ -3581,6 +3581,110 @@ paths:
...
+          in: query
+          description: item paths to get
+          required: true
+          schema:
+            type: array
+            items:
+              type: string
+        - name: preferContent
+          in: query
+          description: return content item in case if it exsits when quering for container
```
sumerjabri: when set to true, return an item instead of a folder if the path can match either

src/main/java/org/craftercms/studio/impl/v2/service/content/internal/ContentServiceInternalImpl.java
```diff
@@ -208,6 +219,59 @@ public DetailedItem getItemById(String siteId, long id, boolean preferContent)
...
+        params.put(SITE_ID, siteId);
+        SiteFeed siteFeed = siteFeedMapper.getSite(params);
+        List<SandboxItem> toRet = new ArrayList<SandboxItem>();
+        List<Item> items = null;
+        if (preferContent) {
+            items = itemDao.getSandboxItemsByPathPreferContent(siteFeed.getId(), paths);
+        } else {
+            items = itemDao.getSandboxItemsByPath(siteFeed.getId(), paths);
+        }
+        if (CollectionUtils.isNotEmpty(items)) {
```
sumerjabri: Seems that the filtering and available actions are the same for byPath and byId, please factor this section out into a common method so we have one place for that code to be maintained.

sumerjabri (changes requested):

sumerjabri (approved):

PR #2415, opened 2021-02-18
Add new event-based bootstrap process

jdrossl:


src/main/java/org/craftercms/studio/impl/v2/utils/spring/context/BootstrapManager.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Central point to control the event-based bootstrap process.
+ * <p>Note: All methods in this class should use the {@link Order} annotation with the default value to ensure the
+ * events are triggered in the right order</p>
+ *
+ * @author joseross
+ * @since 4.0
+ */
+public class BootstrapManager {
```
avasquez614, line 35: Make all log entries info

avasquez614 (changes requested):

sumerjabri (approved):

PR #2419, opened 2021-02-23
[4541] [studio] Bulk publishing not working in Studio with external Aurora Serverless

dejan-brkic:
Fix mysql compatibility issue

https://github.com/craftercms/craftercms/issues/4541


sumerjabri:
@dejan-brkic I'll wait for the final 2 simple statements solution.

PR #2440, opened 2021-03-08
Add delete content-type API

jdrossl:
https://github.com/craftercms/craftercms/issues/1631

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -3172,12 +3172,12 @@ paths:
         '403':
           $ref: '#/components/responses/Forbidden'
 
-  /configuration/content-type/dependencies:
+  /configuration/content-type/usage:
     get:
       tags:
         - configuration
-      summary: Get all depedencies for a given content-type
-      operationId: getContentTypeDepedencies
+      summary: Get all usage of a given content-type
+      operationId: getContentTypeUsage
       parameters:
         - name: siteId
           in: query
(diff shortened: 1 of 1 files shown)


src/main/api/2/studio-api-2.yaml
```diff
@@ -3172,6 +3172,103 @@ paths:
         '403':
           $ref: '#/components/responses/Forbidden'
 
+  /configuration/content-type/dependencies:
```
sumerjabri: Change `dependencies` to `usage`

src/main/api/2/studio-api-2.yaml
```diff
@@ -3172,6 +3172,103 @@ paths:
...
+    get:
+      tags:
+        - configuration
+      summary: Get all depedencies for a given content-type
```
sumerjabri: Get all usage of a given content-type

src/main/api/2/studio-api-2.yaml
```diff
@@ -3172,6 +3172,103 @@ paths:
...
+      operationId: getContentTypeDepedencies
```
sumerjabri: getContentUsage

src/main/api/2/studio-api-2.yaml
```diff
@@ -3172,6 +3172,103 @@ paths:
...
+                      scripts:
+                        type: array
+                        items:
+                          type: string
+                        description: All scripts associated with the given content-type
+                      content:
+                        type: array
+                        items:
+                          type: string
+                        description: All content items associated with the given content-type
```
sumerjabri: All content types instantiated from the given content-type

sumerjabri (changes requested):

sumerjabri (approved):

PR #2447, opened 2021-03-10
Fixes for plugins:

jdrossl:
- Move load file endpoint to webapp
- Fix base url in the configuration
- Add path validations

https://github.com/craftercms/craftercms/issues/4565

### DIFF
--- a/src/main/resources/crafter/studio/studio-config.yaml
+++ b/src/main/resources/crafter/studio/studio-config.yaml
@@ -204,7 +204,7 @@ studio.configuration.management.previewProtectedUrls: >-
   /api/1/site/cache/statistics
 
 # The pattern used to load plugins installed from the marketplaces
-studio.configuration.plugin.base.pattern: /config/studio/plugins/${id}/${type}/${name}
+studio.configuration.plugin.base.pattern: /config/studio/plugins/js/${id}/${type}/${name}
 # List of patterns to invalidate cache when changes are detected in the global repo (separated by commas)
 studio.configuration.cache.global.patterns: /?configuration/.+
 # List of patterns to invalidate cache when changes are detected in the site repo (separated by commas)
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '4.0.0.7'
+  version: '4.0.0.8'
   title: Crafter Studio
   termsOfService: ''
   contact:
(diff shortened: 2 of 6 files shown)


src/main/java/org/craftercms/studio/controller/web/PluginController.java
```diff
@@ -0,0 +1,66 @@
...
+
+import javax.activation.MimetypesFileTypeMap;
+
+/**
+ * Controller that provides the UI with plugin related files
+ * @author joseross
+ * @since 4.0
+ */
+@RestController
+@RequestMapping("/plugin")
```
sumerjabri: Should this be /plugin/1/{API}?

src/main/java/org/craftercms/studio/controller/web/PluginController.java
```diff
@@ -0,0 +1,66 @@
...
+    protected final ConfigurationService configurationService;
+
+    public PluginController(ConfigurationService configurationService) {
+        this.configurationService = configurationService;
+    }
+
+    /**
+     * Returns a single file for a given plugin
+     */
+    @GetMapping("/file")
```
sumerjabri: or the /1/ goes here so it's per API?

src/main/resources/crafter/studio/studio-config.yaml
```diff
@@ -204,7 +204,7 @@ studio.configuration.management.previewProtectedUrls: >-
   /api/1/site/cache/statistics
 
 # The pattern used to load plugins installed from the marketplaces
-studio.configuration.plugin.base.pattern: /config/studio/plugins/${id}/${type}/${name}
+studio.configuration.plugin.base.pattern: /config/studio/plugins/js/${id}/${type}/${name}
```
sumerjabri, line 207: Is this a fix to a prior bug?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/controller/web/PluginController.java
jdrossl: I would suggest `/1/plugin`

src/main/java/org/craftercms/studio/controller/web/PluginController.java
jdrossl: see above

src/main/resources/crafter/studio/studio-config.yaml
jdrossl: Yes, it was introduced with the PR for installing plugins, it's new code that none has used yet

src/main/java/org/craftercms/studio/controller/web/PluginController.java
sumerjabri: Let's do it.

sumerjabri (approved):

PR #2455, opened 2021-03-12
[4556] [studio] Implement Available Actions per the new design

dejan-brkic:
Added blob store available actions filter

https://github.com/craftercms/craftercms/issues/4556

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v2/utils/StudioConfiguration.java
+++ b/src/main/java/org/craftercms/studio/api/v2/utils/StudioConfiguration.java
@@ -353,6 +353,8 @@ public interface StudioConfiguration {
     // Content Service
     String CONTENT_ITEM_EDITABLE_TYPES = "studio.content.item.editableTypes";
 
+    String BLOB_INTERCEPTED_PATHS = "studio.blob.intercepted.paths";
+
     void loadConfig();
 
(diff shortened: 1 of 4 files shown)


sumerjabri (changes requested):
Switch to using isBlob(site, path) method.

sumerjabri (approved):

src/main/java/org/craftercms/studio/api/v2/utils/StudioConfiguration.java
```diff
@@ -353,6 +353,8 @@
     // Content Service
     String CONTENT_ITEM_EDITABLE_TYPES = "studio.content.item.editableTypes";
 
+    String BLOB_INTERCEPTED_PATHS = "studio.blob.intercepted.paths";
```
sumerjabri: Is this needed?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/api/v2/utils/StudioConfiguration.java
dejan-brkic: It should be there and used instead of magic strings

sumerjabri (approved):

PR #2458, opened 2021-03-12
Implement delete content-type API

jdrossl:
https://github.com/craftercms/craftercms/issues/1631

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
@@ -229,7 +229,7 @@ public class SiteServiceImpl implements SiteService {
         auditLog.setPrimaryTargetValue(path);
         auditServiceInternal.insertAuditLog(auditLog);
 
-        itemServiceInternal.persistItemAfterWrite(site, path, commitId, user, Optional.of(true));
+        itemServiceInternal.persistItemAfterWrite(site, path, user, commitId, Optional.of(true));
 
         if (commitId != null) {
             contentRepositoryV2.insertGitLog(site, commitId, 1);
--- a/src/main/resources/org/craftercms/studio/api/v2/dal/ItemDAO.xml
+++ b/src/main/resources/org/craftercms/studio/api/v2/dal/ItemDAO.xml
@@ -602,4 +602,24 @@
             #{path}
         </foreach>
     </update>
+
+    <select id="getContentTypeUsages" resultType="org.craftercms.studio.api.v2.dal.Item">
+        SELECT
+          I.path, I.system_type AS systemType
+        FROM
+          item I
+        INNER JOIN
+          site S
+        ON
+          S.site_id = #{siteId}
+        WHERE
+          I.site_id = S.id
+        AND
+          (
+            (I.system_type IN ('page', 'component') AND I.content_type_id = #{contentType})
+          OR
+            (I.system_type = 'script' AND I.path = #{scriptPath})
+          )
+    </select>
+
 </mapper>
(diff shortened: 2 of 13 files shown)


src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
```diff
@@ -234,7 +234,7 @@ public boolean writeConfiguration(@ValidateStringParam(name = "site") String sit
         auditLog.setPrimaryTargetValue(path);
         auditServiceInternal.insertAuditLog(auditLog);
 
-        itemServiceInternal.persistItemAfterWrite(site, path, commitId, user, Optional.of(true));
+        itemServiceInternal.persistItemAfterWrite(site, path, user, commitId, Optional.of(true));
```
sumerjabri, line 232: Why switch these params?

src/main/resources/org/craftercms/studio/api/v2/dal/ItemDAO.xml
```diff
@@ -580,4 +580,22 @@
         AND s.deleted = 0
         AND i.path = #{path}
     </select>
+
+    <select id="getContentTypeUsages" resultType="org.craftercms.studio.api.v2.dal.Item">
+        SELECT
+          I.path, I.system_type AS systemType
+        FROM
+          item I
+        LEFT JOIN
```
sumerjabri: Switch to INNER JOIN

sumerjabri: Get intersection of two tables on the site_id, and THEN filter by site and type.

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
jdrossl: It's a typo that was unnoticed until now https://github.com/craftercms/studio/blob/730c6658b1a378b7800459630b18df9528bd699e/src/main/java/org/craftercms/studio/api/v2/service/item/internal/ItemServiceInternal.java#L333

src/main/resources/org/craftercms/studio/api/v2/dal/ItemDAO.xml
```diff
@@ -602,4 +602,24 @@
...
+
+    <select id="getContentTypeUsages" resultType="org.craftercms.studio.api.v2.dal.Item">
+        SELECT
+          I.path, I.system_type AS systemType
+        FROM
+          item I
+        INNER JOIN
+          site S
+        ON
+          S.site_id = #{siteId}
```
sumerjabri: Join on S.site_id = I.site_id and then WHERE S.site_id = 'value'

src/main/resources/org/craftercms/studio/api/v2/dal/ItemDAO.xml
```diff
@@ -602,4 +602,24 @@
...
+        WHERE
+          I.site_id = S.id
```
sumerjabri, line 614: S.site_id = 'value'

src/main/resources/org/craftercms/studio/api/v2/dal/ItemDAO.xml
```diff
@@ -602,4 +602,24 @@
...
+        AND
+          (
+            (I.system_type IN ('page', 'component') AND I.content_type_id = #{contentType})
```
sumerjabri, line 619: if #{contentType} is a regex, you should switch it to use SQL patterns and use `like` instead of `=`

Same for the I.path below.

jdrossl: those are simple strings, not regex

sumerjabri (approved):

PR #2462, opened 2021-03-17
Fix name collision for paste operation

jdrossl:
https://github.com/craftercms/craftercms/issues/4555

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
```diff
@@ -1172,7 +1173,7 @@ else if(!fromFileIsIndex && newFileIsIndex) {
         if(adjustOnCollide && contentExists) {
             logger.debug("File already found at path {0}, creating new name", proposedDestPath);
             try {
-                var siblings = contentRepository.getSubtreeItems(site, newPathOnly);
+                var siblings = _contentRepository.getContentChildren(site, newPathOnly);
```
sumerjabri, line 1176: Please switch to explicit typing for this method and future methods.

sumerjabri (changes requested):

sumerjabri (approved):

PR #2464, opened 2021-03-18
[4556] [studio] Implement Available Actions per the new design

dejan-brkic:
Added protected folders configuration

https://github.com/craftercms/craftercms/issues/4556

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/security/SemanticsAvailableActionsResolverImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/security/SemanticsAvailableActionsResolverImpl.java
@@ -16,6 +16,7 @@
 
 package org.craftercms.studio.impl.v2.security;
 
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.craftercms.studio.api.v1.exception.ServiceLayerException;
 import org.craftercms.studio.api.v1.exception.security.UserNotFoundException;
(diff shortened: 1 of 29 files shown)


src/main/java/org/craftercms/studio/impl/v2/security/SemanticsAvailableActionsResolverImpl.java
```diff
@@ -84,6 +92,20 @@ private long applySpecialUseCaseFilters(String username, String siteId, Item ite
             throws ServiceLayerException, UserNotFoundException {
         long result = availableActions;
 
+        if (StringUtils.equals(item.getPath(), HOME_PAGE_PATH)) {
```
sumerjabri, line 95: Why isn't this a pattern like the others?

sumerjabri (changes requested):

sumerjabri (changes requested):
This PR seems to contain other stuff not related to the ticket.

src/main/java/org/craftercms/studio/impl/v2/security/SemanticsAvailableActionsResolverImpl.java
dejan-brkic: hardcoded home page ... it is always on the same path and does not need to be configured. If configured, somebody can remove protection by accident or intentionally

dejan-brkic:
> This PR seems to contain other stuff not related to the ticket.

Merge conflicts and Codacy fixes

sumerjabri (approved):

PR #2519, opened 2021-04-20
[4659] [studio] Publishing black list

dejan-brkic:
Implemented publishing blacklist

https://github.com/craftercms/craftercms/issues/4659

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/deployment/PublishingManagerImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/deployment/PublishingManagerImpl.java
@@ -235,7 +235,7 @@ public class PublishingManagerImpl implements PublishingManager {
                     Arrays.asList(studioConfiguration
                             .getArray(CONFIGURATION_PUBLISHING_BLACKLIST_REGEX, String.class)))) {
                 LOGGER.debug("File " + item.getPath() + " of the site " + site + " will not be published because it " +
-                        "matches black list patterns.");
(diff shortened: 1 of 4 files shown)


src/main/java/org/craftercms/studio/impl/v1/service/deployment/PublishingManagerImpl.java
```diff
@@ -230,6 +231,14 @@ else if (StringUtils.equalsIgnoreCase(LIVE_ENVIRONMENT, item.getEnvironment()) |
                 itemMetadata.setLaunchDate(null);
                 objectMetadataManager.updateObjectMetadata(itemMetadata);
             }
+            if (ContentUtils.matchesPatterns(item.getPath(),
+                    Arrays.asList(studioConfiguration
```
sumerjabri, line 235: Is this going through a cached service?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v1/service/deployment/PublishingManagerImpl.java
dejan-brkic: that is YAML configuration

sumerjabri: It is cached after we looked more deeply at it.

sumerjabri (approved):

PR #2527, opened 2021-04-23
Fixes for blob store

jdrossl:
https://github.com/craftercms/craftercms/issues/4669


sumerjabri (changes requested):
Please switch the data structures to treemap and update GitContentRepo in the two spots we discussed.

sumerjabri (approved):

PR #2556, opened 2021-05-14
[4671] Add screenshot to blueprints

jvega190:
https://github.com/craftercms/craftercms/issues/4671


sumerjabri (changes requested):
Remove `/static-assets`

sumerjabri (approved):

PR #2557, opened 2021-05-17
[4660] [studio] Don't insert .keep files into the database while absorbing content via Git

dejan-brkic:
Ignore .keep files when absorbing content via git

https://github.com/craftercms/craftercms/issues/4660


src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
```diff
@@ -505,6 +505,7 @@ private void processCreatedFiles(String siteId, Map<String, String> createdFiles
         try {
             for (String key : createdFiles.keySet()) {
                 String path = key;
+                logger.error("Path: " + path);
```
sumerjabri: Is this a debug statement?

sumerjabri (changes requested):

sumerjabri (approved):

PR #2560, opened 2021-05-18
[4720] [studio] Create Workflow Management API

dejan-brkic:
API docs updated

https://github.com/craftercms/craftercms/issues/4720

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -5522,6 +5522,167 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /workflow/item_states:
+    get:
+      tags:
+        - workflow
+      summary: Get item workflow states for site
+      description: 'Required permission "get_item_states"'
+      operationId: getItemStates
+      parameters:
+        - name: siteId
+          in: query
+          description: site ID
+          required: true
+          schema:
+            type: string
+        - name: path
+          in: query
+          description: path regex
+          required: false
+          schema:
+            type: string
+        - name: states
+          in: query
+          description: states bitmap mask
+          required: false
+          schema:
+            type: integer
+            format: int64
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+                  items:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/SandboxItem'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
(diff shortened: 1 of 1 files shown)


src/main/api/2/studio-api-2.yaml
```diff
@@ -5522,6 +5522,167 @@ paths:
...
+                  description: true if system processing is to be cleared
+                clearUserLocked:
+                  type: boolean
+                  description: true if user lock is to be cleared
+                live:
+                  type: boolean
+                  description: true if item is to be set as live, otherwise false
+                staged:
+                  type: boolean
+                  description: true if item is to be set as staged, otherwise false
```
rart, line 5608: - Aren't we missing state here too? (i.e. to set the item(s) state)?
- Are all these JSON attributes optional? For example, we may want to just set one of them with out altering the rest.

src/main/api/2/studio-api-2.yaml
```diff
@@ -5522,6 +5522,167 @@ paths:
...
```
rart: - Where/what is the query?
- Weren't we going to have some sort of flag to apply to the entire query vs a given set of items?
- I'm wondering if these two mutating apis could be just one. They share much; we could send either query or item paths, and the backend would apply to either the entire query or the set of paths sent. All json properties - except for siteId - would be optional as they can be conjugated.

src/main/api/2/studio-api-2.yaml
sumerjabri: @dejan-brkic please make the bits optional.

src/main/api/2/studio-api-2.yaml
sumerjabri: @rart the query is the set of paths and bitmap at the beginning.

src/main/api/2/studio-api-2.yaml
```diff
@@ -5522,6 +5522,167 @@ paths:
...
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                siteId: 
+                  type: string
+                  description: site ID
+                items:
+                  type: array
```
sumerjabri, line 5593: Why an Array here, and not at the initial query method?

sumerjabri (changes requested):

src/main/api/2/studio-api-2.yaml
sumerjabri: @rart I don't think merging them is useful. They're quite different.

src/main/api/2/studio-api-2.yaml
sumerjabri: @rart I think we got it all, ping me and I can explain better.

sumerjabri (approved):

PR #2629, opened 2021-06-28
[4797] [studio] Add RetryingOperation Facade in v3.1 like v4

dejan-brkic:
Added Retrying Database Operation Facade

https://github.com/craftercms/craftercms/issues/4797


src/main/java/org/craftercms/studio/api/v2/dal/RetryingDatabaseOperationFacade.java
```diff
@@ -0,0 +1,634 @@
...
+
+    /**
+     * Delete site
+     * @param siteId site identifier
+     * @param state deleted state value
+     * @return
+     */
+    boolean deleteSite(String siteId, String state);
+
+    void updateSiteLastCommitId(Map params);
```
sumerjabri, line 145: Please add JavaDoc where it's missing.

sumerjabri (changes requested):

sumerjabri (approved):

PR #2656, opened 2021-07-15
Fix config cache for missing files

jdrossl:
https://github.com/craftercms/craftercms/issues/4791


jdrossl:
@sumerjabri I turned it to draft because I just found an issue, I'll continue to work on it

sumerjabri (approved):

PR #2663, opened 2021-07-20
[4859] [studio] DetailedItem.{staging,live}.publisher should be a username

dejan-brkic:
Publisher property switched to be username

https://github.com/craftercms/craftercms/issues/4859


sumerjabri (approved):

sumerjabri (changes requested):
Please switch to use DB tables and not Git for the data needed.
Please update docs to indicate that commit_id for staging and live are the versions of the content when it was published and make sure the commit_ids refer to the sandbox.

sumerjabri (approved):

PR #2684, opened 2021-07-30
[4887] [studio] Move the disabled state flag into the state bitmap

dejan-brkic:
Move disabled flag to state bitmap

https://github.com/craftercms/craftercms/issues/4887


sumerjabri (changes requested):
SiteServiceImpl.java seems to have a number of repeated code blocks. Do you think you can factor these out into shared method(s)?

dejan-brkic:
It is not worth it refectoring. It is V1 service, and that is not the only thing that needs refactor/redo

sumerjabri (approved):

PR #2708, opened 2021-08-17
[4771] [studio] Improve API response messages for remote repos

dejan-brkic:
Improved API response messages

https://github.com/craftercms/craftercms/issues/4771


sumerjabri (approved):

jvega190:
@sumerjabri @dejan-brkic while working on https://github.com/craftercms/studio-ui/pull/2521, I realized that the error message when it fails to create a remote (invalid url) is a little bit generic (`Internal system failure: Error while adding remote test (url: /some/local/path) for site 4964`). Is that something that can be improved?

russdanner:
Confirmed

PR #2770, opened 2021-10-12
Remove Plugins:

jdrossl:
- Add new APIs for removing plugins
- Refactor for install plugin to generate a single commit
- Improvements for configuration wiring
- Improvements for dependency scan

https://github.com/craftercms/craftercms/issues/5006

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '4.0.0.22'
+  version: '4.0.0.23'
   title: Crafter Studio
   termsOfService: ''
   contact:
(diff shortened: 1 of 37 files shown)


jdrossl:
Turned to draft because I found an issue that will need some discussion

jdrossl:
XSLT issues will be covered  in https://github.com/craftercms/craftercms/issues/5038

src/main/api/2/studio-api-2.yaml
```diff
@@ -4541,6 +4541,97 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /marketplace/remove:
+    post:
+      tags:
+        - marketplace
+      summary: Remove a plugin from the given site
+      description: 'Required permission "install_plugins"'
```
sumerjabri: Should we have a `remove_plugins` permission?

src/main/api/2/studio-api-2.yaml
```diff
@@ -4541,6 +4541,97 @@ paths:
...
+                    type: array
+                    items:
+                      type: string
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '409':
```
sumerjabri, line 4634: When would this error happen?

src/main/api/2/studio-api-2.yaml
```diff
@@ -4541,6 +4541,97 @@ paths:
...
+                properties:
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '409':
```
sumerjabri, line 4587: Don't we need a 404?

src/main/api/2/studio-api-2.yaml
```diff
@@ -4541,6 +4541,97 @@ paths:
...
+                    $ref: '#/components/schemas/ApiResponse'
+                  items:
+                    type: array
+                    items:
+                      type: string
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '403':
```
sumerjabri, line 4630: Don't we need a 404 (for site not found)?

src/main/java/org/craftercms/studio/impl/v2/service/marketplace/MarketplaceServiceImpl.java
```diff
@@ -93,4 +93,15 @@ public void copyPlugin(String siteId, @ValidateSecurePathParam(name = "path") St
         marketplaceServiceInternal.copyPlugin(siteId, path);
     }
 
+    @Override
+    @HasPermission(type = DefaultPermission.class, action = PERMISSION_INSTALL_PLUGINS)
```
sumerjabri: Should this be updated to `removePlugin` permission?

sumerjabri (changes requested):

src/main/api/2/studio-api-2.yaml
jdrossl: We can add it, but I can't think of a use case in which an user should be able to only install or only remove plugins

src/main/api/2/studio-api-2.yaml
jdrossl: Those are standard errors related to authentication & authorization

src/main/api/2/studio-api-2.yaml
jdrossl: I will add it

src/main/api/2/studio-api-2.yaml
jdrossl: I will add it

src/main/java/org/craftercms/studio/impl/v2/service/marketplace/MarketplaceServiceImpl.java
jdrossl: same as https://github.com/craftercms/studio/pull/2770#discussion_r728460885

src/main/api/2/studio-api-2.yaml
sumerjabri: I can think of the ability to install but not uninstall as a valid configuration. That gives some users the ability to add but not to remove/break this system. Please add it.

src/main/java/org/craftercms/studio/impl/v2/service/marketplace/MarketplaceServiceImpl.java
sumerjabri: Let's add it. See my note earlier.

src/main/api/2/studio-api-2.yaml
jdrossl: which roles should have the permission by default?

sumerjabri: Admin and developer should be defaulted to add/remove.

sumerjabri (approved):

PR #2782, opened 2021-10-25
Fix for JWT timeouts

jdrossl:
https://github.com/craftersoftware/craftercms/issues/283


sumerjabri (changes requested):
Please move the activity tracking to RAM instead of DB for performance concerns.

sumerjabri (approved):

PR #2783, opened 2021-10-28
[5051][next] Switch to write-content API and stop using asset-upload

jvega190:
https://github.com/craftercms/craftercms/issues/5051


jvega190:
@rart this PR is pending, maybe @jdrossl can help approving it

jdrossl (approved):

PR #2792, opened 2021-12-01
Plugins:

jdrossl:
- Copy files only from the expected id folder structure

https://github.com/craftercms/craftercms/issues/5122

### DIFF
--- a/src/main/resources/crafter/studio/studio-config.yaml
+++ b/src/main/resources/crafter/studio/studio-config.yaml
@@ -211,7 +211,7 @@ studio.configuration.management.previewProtectedUrls: >-
   /api/1/site/cache/statistics
 
 # The pattern used to load plugins installed from the marketplaces
-studio.configuration.plugin.base.pattern: /config/studio/plugins/js/${id}/${type}/${name}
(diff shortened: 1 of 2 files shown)


sumerjabri (changes requested):
Switch to using file separator instead of "/"

sumerjabri:
Will merge after the plugins have been updated.

sumerjabri (approved):

src/main/resources/crafter/studio/studio-config.yaml
```diff
@@ -211,7 +211,7 @@ studio.configuration.management.previewProtectedUrls: >-
   /api/1/site/cache/statistics
 
 # The pattern used to load plugins installed from the marketplaces
-studio.configuration.plugin.base.pattern: /config/studio/plugins/js/${id}/${type}/${name}
+studio.configuration.plugin.base.pattern: /config/studio/assets/plugins/${id}/${type}/${name}
```
sumerjabri: @russdanner now is the time to change this name if we want it changed.

sumerjabri: @jdrossl please rename to static-assets

sumerjabri (changes requested):

sumerjabri (approved):

PR #2794, opened 2021-12-02
Update editorial bp - sibebar hamburger z-index

jvega190:


src/main/webapp/repo-bootstrap/global/blueprints/1000_website_editorial/sources/sass/libs/_vars.scss
```diff
@@ -1,6 +1,6 @@
 // Misc.
 	$misc: (
-		z-index-base:		10000
+		z-index-base:		1000
```
jvega190, line 3: @rart this variable is used only in the sidebar, so no other element z-index will be affected

rart (approved):

PR #2798, opened 2021-12-09
[5099] [studio] Publisher stops if any single item in package does not exist in blob store

dejan-brkic:
Publishing from blob store does not block publishing anymore if content is not present

https://github.com/craftercms/craftercms/issues/5099


sumerjabri:
@jdrossl also, why doesn't this happen in v4?

jdrossl (approved):
I'm ok with the fix, but to me it looks like a symptom of a different issue. If you try to publish a .blob and the content doesn't exist in S3 it means something went wrong before the publish

jdrossl:
@sumerjabri in v4 we did a refactor for the exception handling in blob store that could explain why it's different

russdanner:
Confirmed

PR #2825, opened 2022-01-10
Fix for S3:

jdrossl:
- Support copy operation for large files

https://github.com/craftercms/craftercms/issues/5215

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/aws/AwsUtils.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/aws/AwsUtils.java
@@ -29,11 +29,15 @@ import java.io.InputStream;
 import java.util.LinkedList;
 import java.util.List;
 
+import static java.util.stream.Collectors.toList;
+
 public abstract class AwsUtils {
 
     private static final Logger logger = LoggerFactory.getLogger(AwsUtils.class);
 
     public static final int MIN_PART_SIZE = 5 * 1024 * 1024;
+
+    public static final int COPY_PART_SIZE = 1024 * 1024 * 1024;
 
     public static void uploadStream(String inputBucket, String inputKey, AmazonS3 s3Client, int partSize,
                                     String filename, InputStream content) throws AwsException {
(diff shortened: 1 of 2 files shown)


src/main/java/org/craftercms/studio/impl/v1/service/aws/AwsUtils.java
```diff
@@ -29,12 +29,16 @@
 import java.util.LinkedList;
 import java.util.List;
 
+import static java.util.stream.Collectors.toList;
+
 public abstract class AwsUtils {
 
     private static final Logger logger = LoggerFactory.getLogger(AwsUtils.class);
 
     public static final int MIN_PART_SIZE = 5 * 1024 * 1024;
```
avasquez614, line 38: This constant seems to not be used

src/main/java/org/craftercms/studio/impl/v1/service/aws/AwsUtils.java
```diff
@@ -29,12 +29,16 @@
...
 
+    public static final int COPY_PART_SIZE = 1024 * 1024 * 1024;
```
avasquez614, line 40: This constant seems to not be used

src/main/java/org/craftercms/studio/impl/v1/service/aws/AwsUtils.java
```diff
@@ -97,6 +101,65 @@ public static void uploadStream(String inputBucket, String inputKey, AmazonS3 s3
...
     }
 
+    public static void copyFile(String sourceBucket, String sourceKey, String destBucket, String destKey,
+                                int partSize, AmazonS3 client) {
+        logger.debug("Copying file from {}/{} to {}/{}", sourceBucket, sourceKey, destBucket, destKey);
+        GetObjectMetadataRequest metadataRequest = new GetObjectMetadataRequest(sourceBucket, sourceKey);
+        ObjectMetadata metadataResult = client.getObjectMetadata(metadataRequest);
+        long objectSize = metadataResult.getContentLength();
+
+        if (objectSize >= 5L * 1024L * 1024L * 1024L) {
```
avasquez614: Shouldn't this use a constant?

avasquez614 (changes requested):

src/main/java/org/craftercms/studio/impl/v1/service/aws/AwsUtils.java
jdrossl: it is used in other classes

src/main/java/org/craftercms/studio/impl/v1/service/aws/AwsUtils.java
jdrossl: it is used in other classes

avasquez614 (approved):

PR #2833, opened 2022-01-13
[5209] [studio][studio-ui] Lock the items currently in XB edit mode

dejan-brkic:
Update API docs

https://github.com/craftercms/craftercms/issues/5209

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -4,7 +4,7 @@ servers:
 info:
   description: |-
     Crafter Studio API v2
-  version: '4.0.0.24'
+  version: '4.0.0.25'
   title: Crafter Studio
   termsOfService: ''
   contact:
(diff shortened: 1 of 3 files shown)


src/main/api/2/studio-api-2.yaml
```diff
@@ -4249,6 +4249,87 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /content/item_lock_by_path:
```
sumerjabri: Should it be plural, `items_lock*`?

sumerjabri (changes requested):

src/main/api/2/studio-api-2.yaml
dejan-brkic: Requirements:
``` 
[BACKEND] Create API spec (OAS) to manage locks a set of items by path and review design
 item_lock_by_path
```

sumerjabri: switch to items

sumerjabri (changes requested):
update per convo/notes

sumerjabri (approved):

PR #2844, opened 2022-01-21
[5255] [studio] Add Group filtering capability to GET groups API 2

dejan-brkic:
Added keyword parameter for get groups API

https://github.com/craftercms/craftercms/issues/5255


src/main/api/2/studio-api-2.yaml
```diff
@@ -5119,7 +5125,7 @@ paths:
       tags:
         - publishing
       summary: Get available publishing targets for site
-      description: 'Required permission "publish"'
+      description: 'Required permission "publish" '
```
sumerjabri: Why the extra space?

sumerjabri (changes requested):

sumerjabri (approved):

PR #2851, opened 2022-01-25
Fix format for dates on copy

jdrossl:
https://github.com/craftercms/craftercms/issues/5268


sumerjabri (changes requested):
Can this be made a common utility function that's used wherever this is needed to avoid this type of error in the future?

jdrossl:
This is the only place so far where this format is required, but yeah in the future it is possible to reuse it

sumerjabri:
Please centralize it in the utils/helper class as a static method, add Javadoc, and use the shared helper.

sumerjabri (approved):

PR #2872, opened 2022-02-09
[4815] [studio] Design and implement new dashboards

dejan-brkic:
Added new dashboard apis

https://github.com/craftercms/craftercms/issues/4815


src/main/api/2/studio-api-2.yaml
```diff
@@ -5991,119 +5942,195 @@ paths:
...
+      description: 'Required permission "site_member"'
+      operationId: getDashboardContentExpiring
+      parameters:
+        - name: siteId
+          in: query
+          description: site ID
+          required: true
+          schema:
+            type: string
+        - name: dateRange
```
sumerjabri: Split into two fields: dateFrom and dateTo

src/main/api/2/studio-api-2.yaml
```diff
@@ -5991,119 +5942,195 @@ paths:
...
+      parameters:
+        - name: siteId
+          in: query
+          description: site ID
+          required: true
+          schema:
+            type: string
+        - name: unitOfTime
+          in: path
+          description: package ID
```
sumerjabri: Fix desc

sumerjabri (changes requested):

src/main/api/2/studio-api-2.yaml
```diff
@@ -5991,119 +5942,203 @@ paths:
...
+      parameters:
+        - name: siteId
+          in: query
+          description: site ID
+          required: true
+          schema:
+            type: string
+        - name: unitOfTime
+          in: path
+          description: unit of time to display stats for (24 hours, 30 days, 90 days, 1 year)
```
sumerjabri: @dejan-brkic what is the unit of time in? Seconds, hours, or what? Let's pick something we can work with easily. Perhaps day? Or hour? What do you think we can accept and easily turn into a query.

sumerjabri (changes requested):

src/main/api/2/studio-api-2.yaml
sumerjabri: @dejan-brkic let's do day then.

sumerjabri (approved):

PR #2888, opened 2022-03-02
[5347] [studio] Cut-Paste of a page as a child of another page does not update the Sidebar

dejan-brkic:
Fixed cut paste issue caused by wrong parent id reference

https://github.com/craftercms/craftercms/issues/5347


sumerjabri (approved):

src/main/resources/org/craftercms/studio/api/v2/dal/ItemDAO.xml
```diff
@@ -1186,6 +1186,9 @@
             i.state = (i.state | #{onStatesBitMap}) &amp; ~#{offStatesBitMap},
             i.preview_url = REPLACE(i.preview_url, #{oldPreviewUrl}, #{newPreviewUrl}),
             i.previous_path = IFNULL(i.previous_path, i.path)
+        <if test="parentId != null">
```
russdanner, line 1189: What if parent ID is null? Does that create a data issue silently?

russdanner (commented):
I have a question about ItemDAO.xml line 1189. This might generate data issues silently if the parent ID is null

src/main/resources/org/craftercms/studio/api/v2/dal/ItemDAO.xml
dejan-brkic: parent id is null only for site content root

PR #2889, opened 2022-03-02
Event Bus & Websocket:

jdrossl:
- Replace EBus with Spring Events
- Add new websocket & security configuration

https://github.com/craftercms/craftercms/issues/5325

Depends on https://github.com/craftercms/craftercms/pull/5349


src/main/java/org/craftercms/studio/api/v2/event/Person.java
```diff
@@ -0,0 +1,69 @@
...
+
+    protected final String firstName;
+
+    protected final String lastName;
+
+    public Person(String username, String firstName, String lastName) {
+        this.username = username;
+        this.firstName = firstName;
+        this.lastName = lastName;
+    }
```
sumerjabri, line 46: There is more to `Person`, see https://app.swaggerhub.com/apis/craftercms/studio/4.0.0.26#/Person

src/main/java/org/craftercms/studio/api/v2/event/Person.java
```diff
@@ -0,0 +1,69 @@
...
+        return new Person(user.getUsername(), user.getFirstName(), user.getLastName());
+    }
+
+    protected final String username;
+
+    protected final String firstName;
+
+    protected final String lastName;
+
+    public Person(String username, String firstName, String lastName) {
```
sumerjabri, line 42: Should `Person` be a TO used everywhere? We have other usecases.

src/main/java/org/craftercms/studio/api/v2/event/content/ContentEvent.java
```diff
@@ -0,0 +1,68 @@
...
+package org.craftercms.studio.api.v2.event.content;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import org.craftercms.studio.api.v2.event.BroadcastEvent;
+import org.craftercms.studio.api.v2.event.SiteAwareEvent;
+import org.springframework.security.core.Authentication;
+
+/**
+ * @author joseross
+ * @since
```
sumerjabri: Since?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/api/v2/event/Person.java
jdrossl: I think this class doesn't exist in the backend yet

src/main/java/org/craftercms/studio/api/v2/event/Person.java
jdrossl: Yes I added it here temporarily until the real implementation is added and can be used across all services

sumerjabri (approved):

PR #2903, opened 2022-03-15
[5331] [studio] Workflow States: Error in filter regex of Workflow States when adding '*' or '[' characters

dejan-brkic:
Added regex validation

https://github.com/craftercms/craftercms/issues/5331


src/main/java/org/craftercms/studio/controller/rest/v2/WorkflowController.java
```diff
@@ -103,6 +107,16 @@ public ResponseBody getItemStates(@RequestParam(name = REQUEST_PARAM_SITEID) Str
         return responseBody;
     }
 
+    private boolean isPathRegexValid(String pathRegex) {
```
sumerjabri, line 113: Is this going to return a 400 (invalid params) to the REST caller or will it silently fail?

sumerjabri (changes requested):

jdrossl (approved):

sumerjabri (approved):

PR #2920, opened 2022-03-29
[331E] [studio] Unnecessary git log updates

dejan-brkic:
Fixed iissue

https://github.com/craftersoftware/craftercms/issues/331

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/job/StudioSyncRepositoryTask.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/job/StudioSyncRepositoryTask.java
@@ -87,14 +87,15 @@ public class StudioSyncRepositoryTask extends StudioClockTask {
                 if (Objects.nonNull(gl)) {
                     String lastRepoCommitId = contentRepository.getRepoLastCommitId(site);
                     if (StringUtils.equals(lastRepoCommitId, lastProcessedCommit)) {
-                        contentRepository.markGitLogVerifiedProcessed(site, gl.getCommitId());
+                        if (gl.getProcessed() == 0) {
+                            contentRepository.markGitLogVerifiedProcessed(site, gl.getCommitId());
+                        }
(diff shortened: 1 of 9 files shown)


src/main/java/org/craftercms/studio/impl/v2/job/StudioSyncRepositoryTask.java
```diff
@@ -87,14 +87,15 @@ private void syncRepository(String site) throws SiteNotFoundException, IOExcepti
                 if (Objects.nonNull(gl)) {
                     String lastRepoCommitId = contentRepository.getRepoLastCommitId(site);
                     if (StringUtils.equals(lastRepoCommitId, lastProcessedCommit)) {
-                        contentRepository.markGitLogVerifiedProcessed(site, gl.getCommitId());
+                        if (gl.getProcessed() == 0) {
+                            contentRepository.markGitLogVerifiedProcessed(site, gl.getCommitId());
+                        }
                         contentRepository.markGitLogProcessedBeforeMarker(site, gl.getId(), 1);
```
avasquez614: Is there a reason why `contentRepository.markGitLogProcessedBeforeMarker(site, gl.getId(), 1);` is not inside the `if`? This SQL statement seems to be the heaviest from the too.

avasquez614 (changes requested):

src/main/java/org/craftercms/studio/impl/v2/job/StudioSyncRepositoryTask.java
dejan-brkic: that one marks all unprocessed commit ids

avasquez614: yeah but it's still executed even after everything has been marked as processed. Over and over again. Isn't it more efficient to query and see if there are any unprocessed commits before sending an update, if a lot of the time the server could be quiet and there are no commits to process?

avasquez614 (approved):

PR #2923, opened 2022-03-30
Get content tree and get content item performance enhancements

avasquez614:


sumerjabri (approved):

src/main/java/org/craftercms/studio/impl/v2/repository/blob/BlobAwareContentRepository.java
```diff
@@ -322,9 +326,18 @@ public String copyContent(String site, String fromPath, String toPath) {
     public RepositoryItem[] getContentChildren(String site, String path) {
         RepositoryItem[] children = localRepositoryV1.getContentChildren(site, path);
         return Stream.of(children)
-                .peek(item -> item.name = getOriginalPath(item.name))
-                .collect(toList())
-                .toArray(new RepositoryItem[children.length]);
+                     .peek(item -> item.name = getOriginalPath(item.name))
```
jdrossl, line 329: I don't think `peek` is supposed to change the item, maybe we should use `map` instead?

src/main/java/org/craftercms/studio/impl/v2/repository/blob/BlobAwareContentRepository.java
```diff
@@ -322,9 +326,18 @@ public String copyContent(String site, String fromPath, String toPath) {
...
+                     .collect(toList())
+                     .toArray(new RepositoryItem[children.length]);
+    }
+
+    @Override
+    public RepositoryItem[] getContentChildrenShallow(String site, String path) {
+        RepositoryItem[] children = localRepositoryV1.getContentChildrenShallow(site, path);
+        return Stream.of(children)
+                     .peek(item -> item.name = getOriginalPath(item.name))
+                     .collect(toList())
```
jdrossl: For performance I think it could help to use `toArray(...)` skipping the `collect`

jdrossl (changes requested):

jdrossl (approved):

PR #2933, opened 2022-04-04
Fix available actions for top level folders

jdrossl:
https://github.com/craftercms/craftercms/issues/5427
https://github.com/craftercms/craftercms/issues/5425

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v1/constant/StudioConstants.java
+++ b/src/main/java/org/craftercms/studio/api/v1/constant/StudioConstants.java
@@ -156,7 +156,8 @@ public final class StudioConstants {
         "/site/taxonomy",
         "/static-assets",
         "/templates",
-        "/scripts"
+        "/scripts",
+        "/sources"
     );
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/craftercms/studio/api/v1/constant/StudioConstants.java
```diff
@@ -148,7 +150,14 @@
...
 
-    public static final String HOME_PAGE_PATH = "/site/website/index.xml";
+    public static final List<String> TOP_LEVEL_FOLDERS = List.of(
+        "/site/website/index.xml",
+        "/site/components",
+        "/site/taxonomy",
+        "/static-assets",
+        "/templates",
+        "/scripts"
+    );
```
sumerjabri, line 161: In support of the headless BPs and the various VCs, should this list also include /items, /videos, /streams (was it streams?).

Also, we may be standardizing on /source (verify with Roy). So we may want to include that as well.

sumerjabri (changes requested):

sumerjabri (approved):

PR #2975, opened 2022-05-05
API Improvements:

jdrossl:
- Optimize queries for getChildrenByPath & getSandboxItemsByPath
- Remove unused APIs & code

https://github.com/craftercms/craftercms/issues/5448


sumerjabri (approved):

src/main/java/org/craftercms/studio/api/v2/dal/QueryParameterNames.java
```diff
@@ -265,6 +246,8 @@ public final class  QueryParameterNames {
 
     public static final String SYSTEM_TYPE_FOLDER = "systemTypeFolder";
 
+    public static final String PREFER_CONTENT = "preferContent";
```
sumerjabri, line 249: Are we keeping this feature?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/api/v2/dal/QueryParameterNames.java
jdrossl: yes, turns out it is used not only by the UI but the backend too, so I think it's better to keep it for now

sumerjabri (approved):

PR #2981, opened 2022-05-16
Improve deployment error notification mail

jdrossl:
https://github.com/craftercms/craftercms/issues/5590

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/service/notification/NotificationServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/service/notification/NotificationServiceImpl.java
@@ -17,7 +17,6 @@
 package org.craftercms.studio.impl.v2.service.notification;
 
 import java.io.IOException;
-import java.io.PrintWriter;
 import java.io.StringReader;
 import java.io.StringWriter;
 import java.time.ZonedDateTime;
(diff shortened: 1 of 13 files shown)


src/main/java/org/craftercms/studio/impl/v2/service/notification/NotificationServiceImpl.java
```diff
@@ -130,10 +129,13 @@ public void init() {
     @ValidateParams
     public void notifyDeploymentError(@ValidateStringParam(name = "site") final String site, final Throwable throwable,
```
avasquez614, line 131: I think it will be better to pass the whole publish request so we know which is the publish operation (like NEW, MOVE, DELETE, etc).

```
    public void notifyDeploymentError(@ValidateStringParam(name = "site") final String site, final Throwable throwable,
                                      final List<PublishRequest> filesUnableToPublish, final Locale locale) {
        try {
            final NotificationConfigTO notificationConfig = getNotificationConfig(site, locale);
            final Map<String, Object> templateModel = new HashMap<>();
            templateModel.put("deploymentError", ExceptionUtils.getStackTrace(throwable));
            templateModel.put("files", filesUnableToPublish);
            notify(site, notificationConfig.getDeploymentFailureNotifications(), NOTIFICATION_KEY_DEPLOYMENT_ERROR,
                locale, templateModel);
        } catch (Throwable ex) {
            logger.error("Unable to Notify Error", ex);
        }
    }
```

You would need to change the notification FTLs ofc

src/main/java/org/craftercms/studio/impl/v2/service/notification/NotificationServiceImpl.java
```diff
@@ -130,10 +129,13 @@ public void init() {
...
-        try {
+        try (StringWriter stringWriter = new StringWriter();
+             PrintWriter printWriter = new PrintWriter(stringWriter)) {
             final NotificationConfigTO notificationConfig = getNotificationConfig(site);
+            throwable.printStackTrace(printWriter);
+            printWriter.flush();
             final Map<String, Object> templateModel = new HashMap<>();
-            templateModel.put(TEMPLATE_MODEL_DEPLOYMENT_ERROR, throwable);
+            templateModel.put(TEMPLATE_MODEL_DEPLOYMENT_ERROR, stringWriter);
             templateModel.put(TEMPLATE_MODEL_FILES, convertPathsToContent(site, filesUnableToPublish));
```
avasquez614: We shouldn't use `convertPathsToContent` since it's completely unnecessary, could be slow if the list of files is big and could throw an exception if a file is a delete and it has been removed.

avasquez614 (changes requested):

avasquez614 (approved):

PR #2982, opened 2022-05-19
Add Publish All API

jdrossl:
https://github.com/craftercms/craftercms/issues/5591

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/repository/GitContentRepository.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/repository/GitContentRepository.java
@@ -20,14 +20,18 @@ import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.collections4.iterators.ReverseListIterator;
 import org.apache.commons.configuration2.HierarchicalConfiguration;
 import org.apache.commons.configuration2.tree.ImmutableNode;
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.filefilter.*;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.craftercms.commons.crypto.CryptoException;
 import org.craftercms.commons.crypto.TextEncryptor;
+import org.craftercms.commons.lang.RegexUtils;
 import org.craftercms.core.service.ContentStoreService;
 import org.craftercms.core.service.Item;
 import org.craftercms.studio.api.v1.constant.GitRepositories;
 import org.craftercms.studio.api.v1.dal.PublishRequest;
+import org.craftercms.studio.api.v1.dal.SiteFeed;
 import org.craftercms.studio.api.v1.exception.ServiceLayerException;
 import org.craftercms.studio.api.v1.exception.SiteNotFoundException;
 import org.craftercms.studio.api.v1.exception.repository.InvalidRemoteRepositoryCredentialsException;
@@ -64,21 +68,8 @@ import org.craftercms.studio.impl.v2.utils.DateUtils;
 import org.craftercms.studio.impl.v2.utils.RingBuffer;
 import org.craftercms.studio.impl.v2.utils.StudioUtils;
 import org.craftercms.studio.model.rest.content.DetailedItem;
-import org.eclipse.jgit.api.AddCommand;
-import org.eclipse.jgit.api.CheckoutCommand;
-import org.eclipse.jgit.api.CommitCommand;
-import org.eclipse.jgit.api.DeleteBranchCommand;
-import org.eclipse.jgit.api.DiffCommand;
-import org.eclipse.jgit.api.FetchCommand;
-import org.eclipse.jgit.api.Git;
-import org.eclipse.jgit.api.ListBranchCommand;
-import org.eclipse.jgit.api.LogCommand;
-import org.eclipse.jgit.api.MergeCommand;
-import org.eclipse.jgit.api.PullCommand;
-import org.eclipse.jgit.api.RemoteRemoveCommand;
-import org.eclipse.jgit.api.ResetCommand;
-import org.eclipse.jgit.api.RmCommand;
-import org.eclipse.jgit.api.TagCommand;
+import org.eclipse.jgit.api.*;
+import org.eclipse.jgit.api.errors.EmptyCommitException;
 import org.eclipse.jgit.api.errors.GitAPIException;
 import org.eclipse.jgit.api.errors.RefNotFoundException;
 import org.eclipse.jgit.diff.DiffEntry;
@@ -106,6 +97,7 @@ import org.springframework.core.io.Resource;
 import org.springframework.dao.DuplicateKeyException;
 
 import java.io.File;
+import java.io.FileFilter;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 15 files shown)


src/main/java/org/craftercms/studio/impl/v2/repository/GitContentRepository.java
```diff
@@ -1896,6 +1865,104 @@ private boolean createEnvironmentBranch(String siteId, String sandboxBranchName,
...
+                if (CollectionUtils.isNotEmpty(status.getMissing())) {
+                    AddCommand add = git.add().setUpdate(true);
+                    status.getMissing().forEach(add::addFilepattern);
+                    retryingRepositoryOperationFacade.call(add);
+                }
+
+                List<String> patterns = studioConfiguration.getList(CONFIGURATION_PUBLISHING_BLACKLIST_REGEX,
+                                                                             String.class);
+
+                List<String> ignored = status.getUncommittedChanges().stream()
```
sumerjabri: Add a small comment to indicate you're deleting blacklisted files.

src/main/java/org/craftercms/studio/impl/v2/repository/blob/s3/StudioAwsS3BlobStore.java
```diff
@@ -361,4 +356,15 @@ public void initialPublish(String siteId) throws SiteNotFoundException {
         }
     }
 
+    @Override
+    public void publishAll(String siteId, String publishingTarget) {
+        Mapping previewMapping = getMapping(publishingTargetResolver.getPublishingTarget());
+        Mapping targetMapping = getMapping(publishingTarget);
+
+        logger.debug("Performing publish all for site {0} to {1}", siteId, targetMapping);
+        deleteFolder(targetMapping.target, targetMapping.prefix, getClient());
```
sumerjabri: This can be quite expensive. Can we not use AWS Sync mechanics?

sumerjabri (changes requested):
I may have missed it, but where are we setting an audit record of publishAll?
Also, see notes in the files.

src/main/java/org/craftercms/studio/impl/v2/repository/blob/s3/StudioAwsS3BlobStore.java
jdrossl: The Java SDK doesn't provide the `sync` feature from the CLI, the only option is to delete everything and copy it similar to what we do in the git repo

sumerjabri:
We agreed to pursue a manual sync using a change-set.
The publishAll will be split into a stage -> commit/rollback. Between staging and committing, we will sync the blob-store.

src/main/java/org/craftercms/studio/impl/v2/repository/GitContentRepository.java
```diff
@@ -1896,6 +1875,133 @@ private boolean createEnvironmentBranch(String siteId, String sandboxBranchName,
...
+    public RepositoryChanges startPublishAll(String siteId, String publishingTarget) throws ServiceLayerException {
+        // get the published repo
+        SiteFeed site = siteService.getSite(siteId);
+        Repository repo = helper.getRepository(siteId, GitRepositories.PUBLISHED);
+        // if the published repo doesn't exist yet, create it and finish since it will be in sync already
+        if (repo == null) {
+            logger.info("Creating published repository for site {0}", siteId);
+            if (helper.createPublishedRepository(siteId, site.getSandboxBranch())) {
+                // return an empty set so no additional operations are performed
+                return new RepositoryChanges();
```
sumerjabri, line 1894: Where will we handle the blacklist?

src/main/java/org/craftercms/studio/impl/v2/repository/blob/BlobAwareContentRepository.java
```diff
@@ -728,4 +729,44 @@ public void initialPublish(String siteId) throws SiteNotFoundException {
...
+                Set<String> updatedBlobs = findCompatiblePaths(blobStore, gitChanges.getUpdatedPaths());
+                Set<String> deletedBlobs = findCompatiblePaths(blobStore, gitChanges.getDeletedPaths());
+                if (!(updatedBlobs.isEmpty() && deletedBlobs.isEmpty())) {
+                    blobStore.completePublishAll(siteId, publishingTarget,
+                                                 new RepositoryChanges(updatedBlobs, deletedBlobs));
+                }
+            }
+            localRepositoryV2.completePublishAll(siteId, publishingTarget, gitChanges);
+        } catch (Exception e) {
+            logger.error("Error performing initial publish for site {0}", e, siteId);
```
sumerjabri: Where do we unlock the git repo (generalized lock service)?

src/main/java/org/craftercms/studio/impl/v2/repository/GitContentRepository.java
sumerjabri: Will be handled in the `createPublishedRepository` method in a future PR.

sumerjabri (changes requested):

sumerjabri (approved):

src/main/java/org/craftercms/studio/impl/v2/repository/blob/s3/StudioAwsS3BlobStore.java
```diff
@@ -361,4 +358,41 @@ public void initialPublish(String siteId) throws SiteNotFoundException {
...
+    }
+
+    @Override
+    public RepositoryChanges preparePublishAll(String siteId, String publishingTarget) {
+        // this method should not be called
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void completePublishAll(String siteId, String publishingTarget, RepositoryChanges changes) {
```
avasquez614, line 371: Wasn't there a way in S3 to bulk copy and delete?

src/main/java/org/craftercms/studio/impl/v2/service/publish/PublishServiceImpl.java
```diff
@@ -230,6 +229,28 @@ public List<DeploymentHistoryGroup> getDeploymentHistory(String siteId, int days
         return groups;
     }
 
+    @Override
+    @HasPermission(type = CompositePermission.class, action = PERMISSION_PUBLISH)
+    public void publishAll(@ProtectedResourceId(SITE_ID_RESOURCE_ID) String siteId, String publishingTarget)
+            throws ServiceLayerException, UserNotFoundException {
+        publishServiceInternal.publishAll(siteId, publishingTarget);
+
+        SiteFeed siteFeed = siteService.getSite(siteId);
```
avasquez614, line 238: Do we ignore audit records in case of an exception? If not this should be part of a finally

avasquez614 (changes requested):

src/main/java/org/craftercms/studio/impl/v2/repository/blob/s3/StudioAwsS3BlobStore.java
jdrossl: there is a bulk delete, but no bulk copy

src/main/java/org/craftercms/studio/impl/v2/service/publish/PublishServiceImpl.java
jdrossl: as far as I know yeah, audit only records successful operations

src/main/java/org/craftercms/studio/impl/v2/repository/blob/s3/StudioAwsS3BlobStore.java
avasquez614: do you think we would gain a performance improvement if we use the bulk delete?

jdrossl: it might change a lot, probably there will be improvement only if a publish contains a lot of deleted files

avasquez614: @sumerjabri what do you think?

avasquez614 (approved):

PR #2984, opened 2022-05-20
[3.1] Add auto-repair for locked/corrupted git repos

phuongnq:


src/main/java/org/craftercms/studio/api/v2/annotation/RetryingRepositoryOperationAspect.java
```diff
@@ -63,13 +71,15 @@ public void setMaxSleep(int maxSleep) {
...
         Exception lastException;
 
         int numAttempts = 0;
         do {
             numAttempts++;
             try {
-				 // Execute the business code again
+                // Execute the business code again
                 if (numAttempts > 1) {
                     logger.debug("Retrying repository operation attempt " + (numAttempts - 1));
```
sumerjabri: Please don't use '+' in logs. You can use '{0}' and I'll fix it to '{}' when I switch it to slf4j. Or switch the file to slf4j with '{}'.

sumerjabri (changes requested):
This also should be done during startup, not per PR. The repo might be legitimately locked by another external process. We will only do this during startup.

sumerjabri (changes requested):
Please make it clean up on start up only.

src/main/java/org/craftercms/studio/api/v2/annotation/RetryingRepositoryOperationAspect.java
phuongnq: Fixed, Thanks!

phuongnq:
@sumerjabri I implemented on startup for `.git/index.lock` unlock. 
For the case of `.git/index` corrupted, I keep it in each PR since this makes sense that if the index is corrupted, we should fix it immediately and it also aligns with the behavior on deployer: https://github.com/craftercms/deployer/pull/354/files.
Let me know if we don't want to include `.git/index` correction (as this is not a part of the ticket itself)

src/main/java/org/craftercms/studio/api/v2/annotation/RetryingRepositoryOperationAspect.java
```diff
@@ -69,26 +77,35 @@ public Object doRetryingOperation(ProceedingJoinPoint pjp) throws Throwable {
...
                 lastException = ex;
                 if (isRepositoryLocked(ex)) {
-                    logger.debug("Failed to execute " + method.getName() + " after " + numAttempts + " attempts", ex);
-                    if (numAttempts < maxRetries) {
-                        // If the maximum number of retries is not reached, sleep and execute it again
-                        long sleep = (long) (Math.random() * maxSleep);
-                        logger.debug("Git operation failed due to the repository being locked. Will wait for " +
-                                     sleep + " before next retry" + method.getName());
-                        Thread.sleep(sleep);
+                    logger.debug("Failed to execute {} after {} attempts", method.getName(), numAttempts, ex);
```
sumerjabri: Please use '{}' for placeholders instead of just {}

src/main/java/org/craftercms/studio/impl/v2/repository/RepositoryStartupCleanup.java
```diff
@@ -0,0 +1,122 @@
...
+
+    protected SiteService siteService;
+    protected StudioConfiguration studioConfiguration;
+    protected SecurityService securityService;
+    protected UserServiceInternal userServiceInternal;
+    private TextEncryptor encryptor;
+    protected GeneralLockService generalLockService;
+    protected RetryingRepositoryOperationFacade retryingRepositoryOperationFacade;
+
+    public void cleanup() {
```
sumerjabri: rename to unlockRepositories

src/main/java/org/craftercms/studio/impl/v2/repository/RepositoryStartupCleanup.java
```diff
@@ -0,0 +1,122 @@
...
+                logger.debug("Clean up git lock for site {}", siteId);
+                cleanupSite(siteId, SANDBOX, helper);
+                cleanupSite(siteId, PUBLISHED, helper);
+            });
+        } catch (Exception e) {
+            logger.error("Error cleaning up git lock", e);
+        }
+    }
+
+    protected void cleanupSite(String siteId, GitRepositories repository, GitRepositoryHelper helper) {
```
sumerjabri: unlockRepository

src/main/java/org/craftercms/studio/impl/v2/repository/RepositoryStartupCleanup.java
```diff
@@ -0,0 +1,122 @@
...
+        Path repoPath = helper.buildRepoPath(repository, siteId);
+        if (repoPath != null) {
+            String path = repoPath.toAbsolutePath().toString();
+            if (GitUtils.isRepositoryLocked(path)) {
+                try {
+                    logger.warn("The local repository {} is locked, trying to unlock it", path);
+                    GitUtils.unlock(path);
+                    logger.info(".git/index.lock is deleted from local repository {}", path);
```
sumerjabri: '{}'

src/main/resources/crafter/studio/studio-services-context.xml
```diff
@@ -1438,5 +1438,19 @@
         <constructor-arg value="#{studioConfiguration.getProperty('studio.forwarded.headers.enabled')}"/>
     </bean>
 
+    <!-- ////////////////////////////////////// -->
+    <!--               Cleanup                  -->
```
sumerjabri: Unlock Repositories

src/main/java/org/craftercms/studio/impl/v2/repository/RepositoryStartupCleanup.java
sumerjabri: Where are we unlocking the Global repository? (the one that holds the configuration of the whole system not per site)

src/main/java/org/craftercms/studio/impl/v2/repository/RepositoryStartupCleanup.java
```diff
@@ -0,0 +1,122 @@
...
+                cleanupSite(siteId, SANDBOX, helper);
+                cleanupSite(siteId, PUBLISHED, helper);
+            });
+        } catch (Exception e) {
+            logger.error("Error cleaning up git lock", e);
+        }
+    }
+
+    protected void cleanupSite(String siteId, GitRepositories repository, GitRepositoryHelper helper) {
+        Path repoPath = helper.buildRepoPath(repository, siteId);
```
sumerjabri, line 95: Should we use the GeneralizedLockingService to lock the repos before we manipulate them?

sumerjabri (changes requested):

sumerjabri:
> @sumerjabri I implemented on startup for `.git/index.lock` unlock. For the case of `.git/index` corrupted, I keep it in each PR since this makes sense that if the index is corrupted, we should fix it immediately and it also aligns with the behavior on deployer: https://github.com/craftercms/deployer/pull/354/files. Let me know if we don't want to include `.git/index` correction (as this is not a part of the ticket itself)

I'm okay with this. @avasquez614 @jdrossl what do you guys think?

src/main/resources/crafter/studio/studio-services-context.xml
```diff
@@ -1438,5 +1438,19 @@
...
+    <!-- ////////////////////////////////////// -->
+    <bean id="studioRepositoryStartupCleanup" depends-on="databaseInitializer"
+            class="org.craftercms.studio.impl.v2.repository.RepositoryStartupCleanup" init-method="cleanup">
```
jdrossl: Using init methods to perform actions has led to issues in the past because it's very hard to ensure the right order, in 3.1 there is no alternative but in 4.0 we should use Spring's events, [example](https://github.com/craftercms/studio/blob/develop/src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java#L852)

src/main/java/org/craftercms/studio/impl/v2/utils/git/GitUtils.java
```diff
@@ -0,0 +1,62 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Utility methods for Git operations.
+ *
+ * @author Phil Nguyen
+ */
+
+public class GitUtils {
```
jdrossl: This seems like a good candidate to be moved to `commons` since it's shared by Studio & Deployer

@sumerjabri should that be done now? or maybe we just leave a `TODO` for later?

sumerjabri: it would be a good experience for @phuongnq 

@phuongnq please move the Git Utils to the `commons` project and use them in both Deployer and Studio.

src/main/java/org/craftercms/studio/api/v2/annotation/RetryingRepositoryOperationAspect.java
phuongnq: Fixed. Thank you!

src/main/java/org/craftercms/studio/impl/v2/repository/RepositoryStartupCleanup.java
phuongnq: Fixed. Thank you!

src/main/java/org/craftercms/studio/impl/v2/repository/RepositoryStartupCleanup.java
phuongnq: Fixed. Thank you!

src/main/java/org/craftercms/studio/impl/v2/repository/RepositoryStartupCleanup.java
phuongnq: Fixed. Thank you!

src/main/resources/crafter/studio/studio-services-context.xml
phuongnq: Fixed. Thank you!

src/main/java/org/craftercms/studio/impl/v2/repository/RepositoryStartupCleanup.java
phuongnq: Added unlock logic while bootstrapping global repos. Thank you!

src/main/java/org/craftercms/studio/impl/v2/repository/RepositoryStartupCleanup.java
phuongnq: Added lock. Thank you!

src/main/java/org/craftercms/studio/impl/v2/utils/git/GitUtils.java
phuongnq: Moved to `commons`. The following PR must be checked and merged first: https://github.com/craftercms/commons/pull/303

jdrossl (approved):

PR #2985, opened 2022-05-23
Exclude publish blacklist from published repo

phuongnq:
https://github.com/craftercms/craftercms/issues/5598

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
+++ b/src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
@@ -44,7 +44,6 @@ import org.craftercms.studio.api.v1.log.LoggerFactory;
 import org.craftercms.studio.api.v1.service.GeneralLockService;
 import org.craftercms.studio.api.v1.service.security.SecurityService;
 import org.craftercms.studio.api.v2.dal.User;
-import org.craftercms.studio.api.v2.exception.RepositoryLockedException;
 import org.craftercms.studio.api.v2.exception.git.cli.NoChangesToCommitException;
 import org.craftercms.studio.api.v2.repository.RetryingRepositoryOperationFacade;
 import org.craftercms.studio.api.v2.service.security.internal.UserServiceInternal;
@@ -53,6 +52,8 @@ import org.craftercms.studio.impl.v1.repository.git.GitContentRepositoryConstant
 import org.craftercms.studio.impl.v1.repository.git.TreeCopier;
 import org.craftercms.studio.impl.v2.utils.GitUtils;
 import org.craftercms.studio.impl.v2.utils.git.GitCli;
+import org.craftercms.studio.impl.v1.util.ContentUtils;
+import org.eclipse.jgit.api.RmCommand;
 import org.eclipse.jgit.api.AddCommand;
 import org.eclipse.jgit.api.CheckoutCommand;
 import org.eclipse.jgit.api.CloneCommand;
@@ -72,7 +73,6 @@ import org.eclipse.jgit.api.errors.InvalidRemoteException;
 import org.eclipse.jgit.api.errors.JGitInternalException;
 import org.eclipse.jgit.api.errors.TransportException;
 import org.eclipse.jgit.diff.DiffEntry;
-import org.eclipse.jgit.errors.LockFailedException;
 import org.eclipse.jgit.lib.Constants;
 import org.eclipse.jgit.lib.ObjectId;
 import org.eclipse.jgit.lib.ObjectReader;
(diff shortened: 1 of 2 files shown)


src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
```diff
@@ -651,6 +658,37 @@ private boolean checkoutSandboxBranch(String site, Repository sandboxRepo, Strin
...
+                    boolean isMatched = ContentUtils.matchesPatterns(sitePath, patterns);
+                    if (isMatched) {
+                        String gitPath = getGitPath(sitePath);
+                        rmCommand.addFilepattern(gitPath);
+                    }
+                    return FileVisitResult.CONTINUE;
+                }
+            });
+
+            retryingRepositoryOperationFacade.call(rmCommand);
```
jdrossl, line 683: @phuongnq don't we need to add & commit the files after executing the rm?

src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
```diff
@@ -540,6 +546,7 @@ public boolean createPublishedRepository(String siteId, String sandboxBranch) {
             Repository publishedRepo = publishedGit.getRepository();
             optimizeRepository(publishedRepo);
             checkoutSandboxBranch(siteId, publishedRepo, sandboxBranch);
+            removePublishBlackList(publishedRepo);
```
jdrossl, line 549: @phuongnq if the operation is done in the sandbox branch we might need to merge it into the `live` branch

jdrossl (changes requested):

src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
phuongnq: Hi @jdrossl Thanks for the comment. 
It seems like with or without a commit command, the whole files will be added to an "Initial publish." commit anyway. 
But I added a commit command for consistency. Note that when checking out to `live` branch, this commit ID will be merged to an initial commit.

src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
phuongnq: Hi @jdrossl 
Thanks for the comment.
AFAIK, the `live` branch will be created after we created this published repository(?)
If this is the case, the new commit will be automatically merged to the newly created branch switch. Do we have any case that which the `live` branch is created beforehand?

src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
jdrossl: Let's not create an empty commit if it's not needed then, please rollback that change. Thanks

src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
jdrossl: We have to test it with initial publish, regular publish & the new publish all once that branch is merged

src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
phuongnq: Hi @jdrossl The commit is not empty but it will be replaced with an initial commit in `live` branch. I think we can keep it. I will send you a video for the detail

src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
phuongnq: I will send a video for detail

src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
jdrossl: Hi @phuongnq, if that commit is not completely necessary (the changes will be included in the next commit) I think we should avoid it for performance reasons (in large sites each commit can take a long time to complete)

phuongnq: Hi @jdrossl Got it. I removed the commit. Note that blacklist files are only removed in `live` branch. If you switch to `mater` or sandbox branch, it is still there (since we don't commit the change). But I guess this is OK as we only need to remove it in `live`.
diff_live_and_test

jdrossl: yeah, I think they have always been in the other branches

jdrossl (approved):

PR #2990, opened 2022-05-25
Improvements for sync of repos

jdrossl:


src/main/java/org/craftercms/studio/impl/v2/job/StudioClusterPublishedRepoSyncTask.java
```diff
@@ -401,29 +390,51 @@ private Set<String> getAllPublishingEnvironments(String site) {
...
+                    .setCreateBranch(createBranch);
+            if (createBranch) {
+                checkoutCommand.setStartPoint(remoteNode.getGitRemoteName() + "/" + branch);
+            }
+            checkoutCommand.call();
+            CheckoutResult result = checkoutCommand.getResult();
+            // If the checkout didn't throw an exception but still failed
+            if (result.getStatus() != CheckoutResult.Status.OK) {
+                // Throw an exception to trigger the cleanup
+                throw new IllegalStateException("Checkout failed for site " + siteId + " branch " + branch +
```
sumerjabri, line 414: Where would the clean-up happen?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v2/job/StudioClusterPublishedRepoSyncTask.java
jdrossl: In the same catch block, I forgot to add this exception there too

sumerjabri (approved):

PR #2991, opened 2022-05-25
[5604] Make the site cookie domain configurable as base or sub

jvega190:
https://github.com/craftercms/craftercms/issues/5604


rart (commented):
What about all the rest of pages?
image

rart (approved):

PR #2995, opened 2022-05-26
Add auto-repair for locked/corrupted git repos

phuongnq:

The following PR must be checked and merged first: https://github.com/craftercms/commons/pull/304

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
+++ b/src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
@@ -1052,9 +1052,7 @@ public class GitRepositoryHelper implements DisposableBean {
                 String path = globalConfigRepoPath.getParent().toAbsolutePath().toString();
                 if (GitUtils.isRepositoryLocked(path)) {
                     try {
-                        logger.warn("The global repository '{}' is locked, trying to unlock it", path);
                         GitUtils.unlock(path);
-                        logger.info(".git/index.lock is deleted from global repository '{}'", path);
                     } catch (IOException e) {
                         logger.warn("Error unlocking git repository '{}'", path, e);
                     }
--- a/src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
+++ b/src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
@@ -16,6 +16,9 @@
 
 package org.craftercms.studio.api.v2.utils;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import org.apache.commons.collections.MapUtils;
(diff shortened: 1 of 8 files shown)


phuongnq:
@avasquez614 To fix the case git is corrupted, I added a new parameter `Repository` to `RetryingRepositoryOperationFacadeImpl`:

```
@Override
    public <T> T call(Callable<T> gitCommand, Repository repository) throws Exception {
        return gitCommand.call();
    }
```

This is used to track Git repository path while catching the exception. Please let me know if this is OK or if we have a better way to do it.

src/main/java/org/craftercms/studio/impl/v2/repository/RepositoryStartupCleanup.java
```diff
@@ -0,0 +1,114 @@
...
+            unlockSitesRepositories();
+        } catch (Exception e) {
+            logger.error("Error cleaning up git lock", e);
+        }
+    }
+
+    protected void unlockSitesRepositories() {
+        siteService.getAllAvailableSites().forEach(siteId -> {
+            logger.debug("Unlock git lock for site '{}'", siteId);
+            String gitLockKeySandbox = SITE_SANDBOX_REPOSITORY_GIT_LOCK.replaceAll(PATTERN_SITE, siteId);
```
jdrossl: It would be better to use the methods from the helper so that in the future we have this code in a single place, see: https://github.com/craftercms/studio/blob/02a7a6e2df3a49f29ac44efd9039abf8dafedbd9/src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java#L1280

jdrossl (changes requested):

sumerjabri:
@phuongnq yes, you should reset the repo after removal of the index

russdanner:
The manual fix for the currupt repo is to run the following two commands:
```
rm .git/index
git reset --hard
 ```

File: src/main/java/org/craftercms/studio/impl/v2/repository/RepositoryStartupCleanup.java (hunk 1 above)
Comment (phuongnq) [reply src/main/java/org/craftercms/studio/impl/v2/repository/RepositoryStartupCleanup.java]: Fixed. Thank you!

**Comment by phuongnq** [comment]
Updated to run `git reset --hard` after the deletion. Thank you!

**Review by jdrossl — APPROVED** [review]

**Review by avasquez614 — CHANGES_REQUESTED** [review]
- Index removal and repository reset should only happen on startup.
- We should do a `git clean` after the reset so new pending files are removed too.

**Comment by phuongnq** [comment]
Updated to do both unlock and remove corrupted index at startup.
Steps for removing index:

```
git rm .git/index
git reset --hard
git clean --force
```

File: src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java [hunk 2]
```diff
@@ -1047,6 +1048,17 @@ public boolean createGlobalRepo() {
                 }
             } else {
                 logger.info("Detected existing global repository, will not create new one.");
+                // unlock if global repository is locked
+                String path = globalConfigRepoPath.getParent().toAbsolutePath().toString();
+                if (GitUtils.isRepositoryLocked(path)) {
+                    try {
+                        logger.warn("The global repository '{}' is locked, trying to unlock it", path);
+                        GitUtils.unlock(path);
+                        logger.info(".git/index.lock is deleted from global repository '{}'", path);
```
avasquez614: This log entry should be part of the `GitUtils.unlock` method.

avasquez614 (changes requested):

jdrossl (approved):

src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
phuongnq: Fixed. Thank you!

avasquez614 (approved):

sumerjabri (approved):

PR #2997, opened 2022-05-27
[3.1] Fix git corruption on startup only

phuongnq:


jdrossl (approved):

avasquez614 (changes requested):
- Index removal and repository reset should only happen on startup.
- We should do a `git clean` after the reset so new pending files are removed too.

phuongnq:
Updated to do both unlock and remove corrupted index at startup.
Steps for removing index:

```
git rm .git/index
git reset --hard
git clean --force
```

avasquez614 (approved):

PR #2999, opened 2022-05-30
Add policy to disable SVG mime type by default

phuongnq:
https://github.com/craftercms/craftercms/issues/5609

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/service/policy/internal/PolicyServiceInternalImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/service/policy/internal/PolicyServiceInternalImpl.java
@@ -51,6 +51,7 @@ public class PolicyServiceInternalImpl implements PolicyServiceInternal {
     public static final String CONFIG_KEY_STATEMENT = "statement";
     public static final String CONFIG_KEY_PATTERN = "target-path-pattern";
     public static final String CONFIG_KEY_PERMITTED = "permitted";
+    public static final  String CONFIG_KEY_DENIED = "denied";
 
     protected ContentRepository contentRepository;
 
@@ -112,7 +113,7 @@ public class PolicyServiceInternalImpl implements PolicyServiceInternal {
     protected void evaluateAction(HierarchicalConfiguration<?> config, Action action, List<ValidationResult> results,
                                   boolean includeAllowed) {
         try {
-            systemValidator.validate(null, action);
+            systemValidator.validate(null, null, action);
 
             if (config != null) {
                 var statementConfig = config.configurationsAt(CONFIG_KEY_STATEMENT).stream()
@@ -124,7 +125,7 @@ public class PolicyServiceInternalImpl implements PolicyServiceInternal {
 
                     for (var validator : policyValidators) {
                         logger.debug("Evaluating {0} using validator {1}", action, validator.getClass().getSimpleName());
-                        validator.validate(statement.configurationAt(CONFIG_KEY_PERMITTED), action);
+                        validator.validate(statement.configurationAt(CONFIG_KEY_PERMITTED), statement.configurationAt(CONFIG_KEY_DENIED), action);
                     }
                 } else {
                     logger.debug("No statement matches found, skipping {0}", action);
--- a/src/main/java/org/craftercms/studio/impl/v2/service/policy/internal/PolicyServiceInternalImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/service/policy/internal/PolicyServiceInternalImpl.java
@@ -17,8 +17,6 @@ package org.craftercms.studio.impl.v2.service.policy.internal;
 
 import org.apache.commons.configuration2.HierarchicalConfiguration;
 import org.apache.tika.io.FilenameUtils;
-import org.craftercms.studio.api.v1.log.Logger;
-import org.craftercms.studio.api.v1.log.LoggerFactory;
 import org.craftercms.studio.api.v1.repository.ContentRepository;
 import org.craftercms.studio.api.v2.exception.configuration.ConfigurationException;
 import org.craftercms.studio.api.v2.exception.validation.ValidationException;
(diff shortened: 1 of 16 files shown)


jdrossl (commented):
See https://github.com/craftersoftware/marketplace-blueprint/pull/13#discussion_r884902691

src/main/java/org/craftercms/studio/impl/v2/service/policy/internal/PolicyServiceInternalImpl.java
```diff
@@ -51,6 +51,7 @@ public class PolicyServiceInternalImpl implements PolicyServiceInternal {
     public static final String CONFIG_KEY_STATEMENT = "statement";
     public static final String CONFIG_KEY_PATTERN = "target-path-pattern";
     public static final String CONFIG_KEY_PERMITTED = "permitted";
+    public static final  String CONFIG_KEY_DENIED = "denied";
```
jdrossl, line 55: @sumerjabri @avasquez614 please confirm if the name is ok

src/main/java/org/craftercms/studio/impl/v2/service/policy/validators/MimeTypePolicyValidator.java
```diff
@@ -40,15 +41,34 @@ public class MimeTypePolicyValidator implements PolicyValidator {
...
     @Override
-    public void validate(HierarchicalConfiguration<?> config, Action action) throws ValidationException {
-        if (config.containsKey(CONFIG_KEY_MIME_TYPES) && !isEmpty(FilenameUtils.getExtension(action.getTarget()))) {
+    public void validate(HierarchicalConfiguration<?> permittedConfig, HierarchicalConfiguration<?> deniedConfig, Action action) throws ValidationException {
+        if (isEmpty(FilenameUtils.getExtension(action.getTarget()))) {
+            logger.debug("Skipping folder '{}'", action.getTarget());
+            return;
+        }
+
+        var isContainPermitConfig = permittedConfig.containsKey(CONFIG_KEY_MIME_TYPES);
```
jdrossl: @phuongnq please change the name to `containsPermittedConfig`, the `is` prefix is usually used only for getter methods

src/main/webapp/repo-bootstrap/global/configuration/samples/sample-site-policy-config.xml
```diff
@@ -44,12 +44,16 @@
...
 
     <!-- Example: only allow images of less than 1 MB -->
+    <!-- disable svg files -->
     <statement>
         <target-path-pattern>/static-assets/images/.*</target-path-pattern>
         <permitted>
             <maximum-file-size>1000000</maximum-file-size>
             <mime-types>image/*</mime-types>
         </permitted>
+        <denied>
```
jdrossl, line 59: @phuongnq changes in samples need to be added to the UM, see [here](https://github.com/craftercms/studio/blob/develop/src/main/resources/crafter/studio/upgrade/pipelines.yaml#L365) for an example and let me know if you have questions about this

src/main/java/org/craftercms/studio/impl/v2/service/policy/validators/ContentTypePolicyValidator.java
```diff
@@ -60,4 +59,8 @@ public void validate(HierarchicalConfiguration<?> config, Action action) throws
         }
     }
 
+    @Override
+    public void validate(HierarchicalConfiguration<?> permittedConfig, HierarchicalConfiguration<?> deniedConfig, Action action) throws ValidationException {
```
jdrossl: @phuongnq it would be better to implement the `denied` for all existing validators right away, not just for the mime-type

jdrossl (changes requested):
@phuongnq the policy service is one of the few that has [unit tests](https://github.com/craftercms/studio/blob/develop/src/test/java/org/craftercms/studio/impl/v2/service/policy/internal/PolicyServiceInternalImplTest.java#L62), please make sure those are updated with the new changes too

src/main/java/org/craftercms/studio/impl/v2/service/policy/validators/ContentTypePolicyValidator.java
phuongnq: For `content-type` it is straightforward to have `denined`.
But how about file size and `path`?

```
<minimum-file-size/> (Minimum size of the file in bytes)
<maximum-file-size/> (Maximum size of the file in bytes)
```

These min/max sizes seem not a good term for denied, what could be the rule?
The same for `path`

```
<path>
          <source-regex/> (Regular expression to validate the full path of the file)
          <target-regex caseTransform="lowercase"/> (Expression to transform the full path of the file)
        </path>
```

jdrossl: you are right, it only makes sense for the validators that perform a comparison against a list, please go ahead and add it for the content-type and then we can document the ones that do support it

src/main/java/org/craftercms/studio/impl/v2/service/policy/internal/PolicyServiceInternalImpl.java
phuongnq: > @phuongnq the policy service is one of the few that has [unit tests](https://github.com/craftercms/studio/blob/develop/src/test/java/org/craftercms/studio/impl/v2/service/policy/internal/PolicyServiceInternalImplTest.java#L62), please make sure those are updated with the new changes too

Updated test case. Thanks!

src/main/java/org/craftercms/studio/impl/v2/service/policy/validators/MimeTypePolicyValidator.java
phuongnq: Updated. Thanks!

src/main/webapp/repo-bootstrap/global/configuration/samples/sample-site-policy-config.xml
phuongnq: Updated UM. Thanks! (Let me know if I missed any configuration)

src/main/java/org/craftercms/studio/impl/v2/service/policy/validators/ContentTypePolicyValidator.java
phuongnq: Updated to have `content-type` with `denied` term. Thanks!

jdrossl (approved):

avasquez614 (approved):

PR #3001, opened 2022-05-31
Remove unnecessary synchronized block and use lock service

phuongnq:
https://github.com/craftersoftware/craftercms/issues/332

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
@@ -143,8 +143,6 @@ import static org.craftercms.studio.api.v1.constant.StudioConstants.PATTERN_SITE
 import static org.craftercms.studio.api.v1.constant.StudioConstants.PATTERN_TO_PATH;
 import static org.craftercms.studio.api.v1.constant.StudioConstants.REPO_COMMIT_MESSAGE_PATH_VAR;
 import static org.craftercms.studio.api.v1.constant.StudioConstants.REPO_COMMIT_MESSAGE_USERNAME_VAR;
-import static org.craftercms.studio.api.v1.constant.StudioConstants.SITE_PUBLISHED_REPOSITORY_GIT_LOCK;
-import static org.craftercms.studio.api.v1.constant.StudioConstants.SITE_SANDBOX_REPOSITORY_GIT_LOCK;
 import static org.craftercms.studio.api.v2.utils.StudioConfiguration.BLUE_PRINTS_PATH;
 import static org.craftercms.studio.api.v2.utils.StudioConfiguration.BOOTSTRAP_REPO;
 import static org.craftercms.studio.api.v2.utils.StudioConfiguration.CLUSTERING_NODE_REGISTRATION;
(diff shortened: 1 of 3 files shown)


src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
```diff
@@ -261,11 +261,12 @@ public InputStream getContent(String site, String path) throws ContentNotFoundEx
     public String writeContent(String site, String path, InputStream content) {
         // Write content to git and commit it
         String commitId = null;
-        String gitLockKey = SITE_SANDBOX_REPOSITORY_GIT_LOCK.replaceAll(PATTERN_SITE, site);
+        String gitLockKey = StringUtils.isEmpty(site) ?
+                                GLOBAL_REPOSITORY_GIT_LOCK :
+                                SITE_SANDBOX_REPOSITORY_GIT_LOCK.replaceAll(PATTERN_SITE, site);
```
jdrossl: @phuongnq this would be a good time to use the methods from the helper to get the keys where possible

jdrossl (changes requested):

src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
sumerjabri: `getRepoCacheKey`

sumerjabri (changes requested):
Please use `getRepoCacheKey`

src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
phuongnq: updated to use `getSandboxRepoLockKey` and `getPublishedRepoLockKey` methods

phuongnq:
> Please use `getRepoCacheKey`

Hi @sumerjabri , where do we need to use `getRepoCacheKey`? I believe this method is used to get cached repo. We have already used `helper.getRepository` method, which use `getRepoCacheKey` internal.

sumerjabri (approved):

sumerjabri:
> > Please use `getRepoCacheKey`
> 
> Hi @sumerjabri , where do we need to use `getRepoCacheKey`? I believe this method is used to get cached repo. We have already used `helper.getRepository` method, which use `getRepoCacheKey` internal.

You're right, all good now with getRepository

src/main/java/org/craftercms/studio/impl/v2/repository/GitContentRepository.java
```diff
@@ -512,25 +511,26 @@ public List<RepoOperation> getOperationsFromDelta(String site, String commitIdFr
     @Override
     public String getRepoFirstCommitId(final String site) {
         String toReturn = EMPTY;
+        String gitLockKey = StringUtils.isEmpty(site) ?
```
avasquez614:
```
StringUtils.isEmpty(site) ?
                                GLOBAL_REPOSITORY_GIT_LOCK :
                                helper.getSandboxRepoLockKey(site);
```

Seems to be repeated everywhere. Can we just change `getSandboxRepoLockKey` to receive an additional parameter, a boolean `ifSiteEmptyUseGlobalRepoLockKey`, and add this check in there?

avasquez614 (changes requested):

jdrossl (approved):

src/main/java/org/craftercms/studio/impl/v2/repository/GitContentRepository.java
phuongnq: Fixed. Thank you!

avasquez614 (approved):

PR #3002, opened 2022-05-31
Add Move content event with source path

phuongnq:
https://github.com/craftercms/craftercms/issues/5559

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v2/event/content/MoveContentEvent.java
+++ b/src/main/java/org/craftercms/studio/api/v2/event/content/MoveContentEvent.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright (C) 2007-2022 Crafter Software Corporation. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as published by
+ * the Free Software Foundation.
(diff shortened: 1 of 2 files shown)


src/main/java/org/craftercms/studio/api/v2/event/content/MoveContentEvent.java
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.security.core.Authentication;
+
+/**
+ * Event triggered when there is a content move change
+ *
+ * @author Phil Nguyen
+ * @since 4.0.1
+ */
+
+public class MoveContentEvent extends SiteAwareEvent implements BroadcastEvent {
```
jdrossl: @phuongnq this class should inherit from `ContentEvent` the idea is to eventually have specific classes for all events and turn `ContentEvent` into an abstract class

jdrossl (changes requested):

src/main/java/org/craftercms/studio/api/v2/event/content/MoveContentEvent.java
phuongnq: Fixed to inherit from `ContentEvent`. Thank you!

jdrossl (approved):

PR #3006, opened 2022-06-01
Fix Publish All for sites without published repo

jdrossl:
https://github.com/craftercms/craftercms/issues/5591

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/repository/GitContentRepository.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/repository/GitContentRepository.java
@@ -1148,6 +1148,10 @@ public class GitContentRepository implements ContentRepository, DeploymentHistor
         retryingRepositoryOperationFacade.call(deleteCommand);
     }
 
+    protected boolean branchExists(Repository repo, String branch) throws IOException {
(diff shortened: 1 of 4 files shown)


src/main/java/org/craftercms/studio/impl/v2/repository/GitContentRepository.java
```diff
@@ -1795,15 +1796,9 @@ public RepositoryChanges preparePublishAll(String siteId, String publishingTarge
...
+        // if the published repo doesn't exist yet, fail
         if (repo == null) {
-            logger.info("Creating published repository for site {0}", siteId);
-            if (helper.createPublishedRepository(siteId, site.getSandboxBranch())) {
-                // return an empty set so no additional operations are performed
-                return new RepositoryChanges();
-            } else {
-                throw new ServiceLayerException("Error creating published repository for site " + siteId);
-            }
+            throw new PublishedRepositoryNotFoundException("Published repository not found for site " + siteId);
```
avasquez614, line 1805: We should also check if the `live` branch exists, and fail with the same exception or a similar one.

avasquez614 (changes requested):

avasquez614 (approved):

PR #3009, opened 2022-06-06
Prevent creating new page on existing folder

phuongnq:
https://github.com/craftercms/craftercms/issues/5593

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v2/exception/content/ContentExistException.java
+++ b/src/main/java/org/craftercms/studio/api/v2/exception/content/ContentExistException.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) 2007-2022 Crafter Software Corporation. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as published by
+ * the Free Software Foundation.
+ *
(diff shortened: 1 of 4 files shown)


sumerjabri:
CC @avasquez614

src/main/java/org/craftercms/studio/api/v2/exception/content/FolderExistException.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.craftercms.studio.api.v2.exception.content;
+
+import org.craftercms.studio.api.v1.exception.ServiceLayerException;
+
+public class FolderExistException extends ServiceLayerException {
```
jdrossl: @phuongnq any time a new exception is created it should be added to the exception handlers in API 2, if not required for now at least add a TODO here to remember it later

jdrossl (changes requested):

src/main/java/org/craftercms/studio/api/v2/exception/content/FolderExistException.java
phuongnq: Added to exception handlers in API 2. Thank you!

jdrossl (approved):

PR #3022, opened 2022-06-15
Preventing search params from exceeding the max result window (offset + limit) of 10000

jmendeza:
https://github.com/craftercms/craftercms/issues/5622


sumerjabri:
CC @russdanner @avasquez6 14 heads up. We're capping the result window to 10K. See https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html

For search results bigger than 10K, it's best to use the methods recommended in the article referenced above.

jdrossl (approved):

PR #3029, opened 2022-06-22
Bugfix/5639

jmendeza:
https://github.com/craftercms/craftercms/issues/5639
Preventing content from being pasted across root folders (e.g.: from scripts into templates)
Handle cases where either source or target content does not exist
Adding unit tests


phuongnq:
@sumerjabri @jmendeza Do we have to worry about the case where some assets are manually added to `/site/website` via git command and then being pasted to a page?
Similarly, there is a case if we manually create 2 new folders from root (Ex: app1, app2) and then add a page to `app1`, a static asset to `app2`. 

In these cases, pasting static assets to a page is possible since we are only invalid paste if the source & target top-level paths are different.

sumerjabri:
@phuongnq good points. We won't be worried about the `git` path, it can't be helped. If someone wants to create a mess, they always can.

We also won't be monitoring the top-level folders created outside of our convention, it's really up to the Information Architect if they want to do something different. It'll be weird, but can't be helped.

sumerjabri (approved):

PR #3030, opened 2022-06-23
Adding logging to validate token process. Method refactor to make it …

jmendeza:
Adding logging to validate token process.
Method refactor to make it more testable.
Added unit tests

https://github.com/craftercms/craftercms/issues/5648

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/service/security/UserServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/service/security/UserServiceImpl.java
@@ -524,7 +524,7 @@ public class UserServiceImpl implements UserService {
             UserExternallyManagedException, ServiceLayerException {
         String decryptedToken = decryptToken(token);
         if (StringUtils.isEmpty(decryptedToken)) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/craftercms/studio/impl/v2/service/security/UserServiceImpl.java
```diff
@@ -552,30 +522,50 @@ public User setPassword(String token, String newPassword) throws UserNotFoundExc
...
+        }
+
+        if (userProfile.isExternallyManaged()) {
+            logger.debug("Invalid Token. User '{}' is externally managed", username);
+            throw new UserExternallyManagedException();
+        }
+
+        String studioId = tokenElements.nextToken();
+        if (!StringUtils.equals(studioId, instanceService.getInstanceId())) {
+            logger.debug("Invalid Token. Token instance ID: '{}' does not match current value: '{}'",
```
sumerjabri: Isn't this token forgery? If so, would it make sense to log it as a WARN?

sumerjabri (changes requested):

sumerjabri (approved):

PR #3045, opened 2022-07-05
Securing /monitoring/log requests

jmendeza:
https://github.com/craftersoftware/craftercms/issues/351
Adding ProxyService to secure /monitoring/log requests. Adding log_monitor permission to developer role


src/main/java/org/craftercms/studio/impl/v2/service/proxy/ProxyServiceImpl.java
```diff
@@ -0,0 +1,144 @@
...
+
+import static org.craftercms.studio.api.v2.utils.StudioConfiguration.CONFIGURATION_MANAGEMENT_PREVIEW_AUTHORIZATION_TOKEN;
+import static org.craftercms.studio.api.v2.utils.StudioConfiguration.CONFIGURATION_MANAGEMENT_PREVIEW_PROTECTED_URLS;
+import static org.craftercms.studio.controller.rest.v2.RequestMappingConstants.PROXY_ENGINE;
+import static org.craftercms.studio.permissions.PermissionResolverImpl.SITE_ID_RESOURCE_ID;
+
+/**
+ * Default implementation for {@link ProxyService}.
+ */
+public class ProxyServiceImpl implements ProxyService {
```
avasquez614, line 39: Shouldn't there be an internal service for this class, which actually contains the business logic? Not sure if we're continuing that pattern.

src/main/resources/crafter/studio/upgrade/pipelines.yaml
```diff
@@ -666,6 +666,12 @@ configurations:
             - type: xsltFileUpgrader
               template: crafter/studio/upgrade/4.0.x/config/permission-mappings-config/permission-mappings-config-v4.0.6.xslt
               commitDetails: Add content_search permissions for all users
+        - currentVersion: 4.0.6
+          nextVersion: 4.0.7
+          operations:
+            - type: xsltFileUpgrader
+              template: crafter/studio/upgrade/4.0.x/config/permission-mappings-config/permission-mappings-config-v4.0.7.xslt
+              commitDetails: Add log_monitor permission for developer role
```
avasquez614: Not sure if `log_monitor` is a good permission name. What about `engin_view_log`?

avasquez614 (changes requested):

src/main/resources/crafter/studio/upgrade/pipelines.yaml
sumerjabri: or `view_logs`

jmendeza: We are using the same permission for studio logs and engine site logs. In that case `view_logs` would be more appropriate
@sumerjabri @avasquez614

sumerjabri: let's do `view_logs` for all then if that works.

sumerjabri (approved):

avasquez614 (approved):

sumerjabri:
@jmendeza please fix conflicts

PR #3060, opened 2022-07-19
Prevent 'already unlock' error when trying to unlock content created with the 'Save' (not close) option

jmendeza:
https://github.com/craftercms/craftercms/issues/5684
Prevent 'already unlock' error when trying to unlock content created with the 'Save' (not close) option


src/main/java/org/craftercms/studio/impl/v2/service/item/internal/ItemServiceInternalImpl.java
```diff
@@ -413,6 +413,7 @@ public void persistItemAfterCreate(String siteId, String path, String username,
             if (unlock) {
                 item.setState(ItemState.savedAndClosed(item.getState()));
             } else {
+                item.setLockedBy(userObj.getId());
```
sumerjabri, line 416: Shouldn't the current user be the user holding the lock already? Why set it again?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/item/internal/ItemServiceInternalImpl.java
jmendeza: When creating an item with "Save" (unlock=false), content is being created as locked since the beginning, but `lockedBy` is not assigned. `lockedBy` being null is what is causing issues when trying to unlock afterward.

sumerjabri (approved):

PR #3117, opened 2022-10-21
Trigger PublishEvent instead of WorkflowEvent on initial publish

jmendeza:
https://github.com/craftercms/craftercms/issues/5770
Trigger PublishEvent instead of WorkflowEvent on initial publish


sumerjabri:
Pending 4.0.2 release.

sumerjabri (approved):

PR #3120, opened 2022-10-27
Audit log API updates

jmendeza:
https://github.com/craftercms/craftercms/issues/5773
Audit log API updates


src/main/java/org/craftercms/studio/api/v2/service/audit/AuditService.java
```diff
@@ -93,17 +66,18 @@ List<AuditLog> getAuditLog(String siteId, String siteName, int offset, int limit
...
 
     /**
      * Get audit log entry by id
      *
      * @param auditLogId audit log id
+     * @param siteId site ID
      * @return audit log entry
      */
-    AuditLog getAuditLogEntry(long auditLogId);
+    AuditLog getAuditLogEntry(long auditLogId, String siteId);
```
sumerjabri: Wouldn't it be better to lead with the `siteId` then the `auditLogId` within it?

src/main/java/org/craftercms/studio/api/v2/service/audit/AuditService.java
```diff
@@ -93,17 +66,18 @@ List<AuditLog> getAuditLog(String siteId, String siteName, int offset, int limit
...
+    int getAuditLogTotal(String siteId, String user, List<String> operations,
                                     boolean includeParameters, ZonedDateTime dateFrom, ZonedDateTime dateTo,
-                                    String target, String origin, String clusterNodeId);
+                                    String target, String origin, String clusterNodeId) throws SiteNotFoundException;
 
     /**
      * Get audit log entry by id
      *
      * @param auditLogId audit log id
+     * @param siteId site ID
```
sumerjabri: Perhaps we also document that this can be `null` which means all sites and admin activities.

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/api/v2/service/audit/AuditService.java
jmendeza: Updated

src/main/java/org/craftercms/studio/api/v2/service/audit/AuditService.java
jmendeza: Updated on other methods retrieving filtered (by site) log entry lists

src/main/java/org/craftercms/studio/api/v2/service/audit/internal/AuditServiceInternal.java
```diff
@@ -68,12 +67,25 @@ List<AuditLog> getAuditLogForSite(String site, int offset, int limit, String use
...
+    List<AuditLog> getAuditLog(String siteId, int offset, int limit, String user,
                                List<String> operations, boolean includeParameters, ZonedDateTime dateFrom,
                                ZonedDateTime dateTo, String target, String origin, String clusterNodeId, String sort,
                                String order);
 
-    int getAuditLogTotal(String siteId, String siteName, String user, List<String> operations,
+    /**
+     * Get the audit log entry count given the provided filter parameters
+     *
+     * @param siteId            site ID     * @param user filter by user
```
sumerjabri: typo (missing CRLF)

sumerjabri (changes requested):

sumerjabri (approved):

PR #3121, opened 2022-10-28
Return 400 when trying to cut-paste into the same location

jmendeza:
https://github.com/craftercms/craftercms/issues/5774
Return 400 when trying to cut-paste into the same location


src/main/api/2/studio-api-2.yaml
```diff
@@ -7398,6 +7398,7 @@ components:
           * `CODE:` 7001, `MESSAGE:` Content already exists, `REMEDIAL ACTION:` Edit the existing item or delete it before creating it again
           * `CODE:` 7002, `MESSAGE:` Content already locked, `REMEDIAL ACTION:` The user that locked the item or the administrator must unlock the item first
           * `CODE:` 7003, `MESSAGE:` Content already unlocked, `REMEDIAL ACTION:` The item is already unlocked
+          * `CODE:` 7004, `MESSAGE:` Content move destination is invalid, `REMEDIAL ACTION:` Check API and make sure you're sending the correct parameters
```
sumerjabri: If this message is displayed to the user (which the UI will probably do), it won't be helpful.

sumerjabri: Message: Paste destination folder is invalid (did you paste the item onto itself?)
Remedial: Try pasting the content to a different folder

sumerjabri (changes requested):

sumerjabri (approved):

PR #3135, opened 2022-11-29
Check when updating externally managed users

phuongnq:
https://github.com/craftersoftware/craftercms/issues/460

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/service/security/UserServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/service/security/UserServiceImpl.java
@@ -147,7 +147,9 @@ public class UserServiceImpl implements UserService {
 
     @Override
     @HasPermission(type = DefaultPermission.class, action = PERMISSION_UPDATE_USERS)
-    public void updateUser(User user) throws ServiceLayerException, UserNotFoundException, AuthenticationException {
+    public void updateUser(User user) throws ServiceLayerException, UserNotFoundException, AuthenticationException, UserExternallyManagedException {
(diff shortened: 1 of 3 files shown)


sumerjabri (approved):

src/main/java/org/craftercms/studio/impl/v2/service/security/UserServiceImpl.java
```diff
@@ -147,7 +147,10 @@ public User createUser(User user) throws UserAlreadyExistsException, ServiceLaye
 
     @Override
     @HasPermission(type = DefaultPermission.class, action = PERMISSION_UPDATE_USERS)
-    public void updateUser(User user) throws ServiceLayerException, UserNotFoundException, AuthenticationException {
+    public void updateUser(User user) throws ServiceLayerException, UserNotFoundException, AuthenticationException, UserExternallyManagedException {
+        checkExternallyManagedUsers(new ArrayList<>(Arrays.asList(user.getId())),
```
jmendeza: Can we simplify this to something like:
```
checkExternallyManagedUsers(Arrays.asList(user.getId()),
                Arrays.asList(user.getUsername()));
```
?

jmendeza (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/security/UserServiceImpl.java
phuongnq: Yes, fixed this. Thank you!

jmendeza (approved):

PR #3144, opened 2022-12-21
Add SiteNotFoundException and ConfigurationProfileNotFoundException to webdav and s3 services

phuongnq:
https://github.com/craftercms/craftercms/issues/5798

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/service/webdav/WebDavServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/service/webdav/WebDavServiceImpl.java
@@ -97,10 +97,7 @@ public class WebDavServiceImpl implements WebDavService {
                                  @ValidateStringParam(name = "profileId") final String profileId,
                                  @ValidateStringParam(name = "path") final String path,
                                  @ValidateStringParam(name = "type") final String type) throws WebDavException, SiteNotFoundException {
-        if (!siteService.exists(siteId)) {
-            throw new SiteNotFoundException(siteId);
-        }
-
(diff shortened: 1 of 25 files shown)


src/main/java/org/craftercms/studio/impl/v2/service/webdav/WebDavServiceImpl.java
```diff
@@ -89,7 +96,11 @@ public List<WebDavItem> list(@ValidateStringParam(name = "siteId")
                                  @ProtectedResourceId("siteId") final String siteId,
                                  @ValidateStringParam(name = "profileId") final String profileId,
                                  @ValidateStringParam(name = "path") final String path,
-                                 @ValidateStringParam(name = "type") final String type) throws WebDavException {
+                                 @ValidateStringParam(name = "type") final String type) throws WebDavException, SiteNotFoundException {
+        if (!siteService.exists(siteId)) {
```
jmendeza: Let's use 
```
siteService.checkSiteExists(siteId)
```

jmendeza (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/webdav/WebDavServiceImpl.java
phuongnq: Fixed this. I also updated the PR to add `ConfigurationProfileNotFoundException`. Please find the detail [here](https://github.com/craftercms/craftercms/issues/5798#issuecomment-1367190473).

jmendeza (approved):

PR #3151, opened 2023-01-03
Sort groups by name by default

phuongnq:
https://github.com/craftercms/craftercms/issues/5742


jmendeza (changes requested):
Let's use @ValidateParams for this. This is probably a common scenario, so we can add the regex to commons and enforce it by using @EsapiValidatedParam
Please take a look at my comment on https://github.com/craftersoftware/craftercms/issues/417 about what needs to be configured for that.

phuongnq:
I have updated PR to use Esapi validator.

phuongnq:
Commons code in this PR https://github.com/craftercms/commons/pull/320

jmendeza (approved):

PR #3156, opened 2023-01-04
Respond 404 for configuration API if content not found

phuongnq:
https://github.com/craftercms/craftercms/issues/5873

### DIFF
--- a/src/main/java/org/craftercms/studio/controller/rest/v2/ConfigurationController.java
+++ b/src/main/java/org/craftercms/studio/controller/rest/v2/ConfigurationController.java
@@ -45,6 +45,7 @@ import javax.validation.Valid;
 import java.beans.ConstructorProperties;
 import java.io.InputStream;
 
+import static java.lang.String.format;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.craftercms.commons.validation.annotations.param.EsapiValidationType.*;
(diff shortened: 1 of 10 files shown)


src/main/java/org/craftercms/studio/controller/rest/v2/ConfigurationController.java
```diff
@@ -85,13 +86,19 @@ public ResponseBody clearCache(@EsapiValidatedParam(type = SITE_ID) @RequestPara
...
+                                         @EsapiValidatedParam(type = ALPHANUMERIC, notNull = false, notEmpty = false, notBlank = false) @RequestParam(name = "environment", required = false) String environment)
+            throws ContentNotFoundException {
         final String content;
         if (StringUtils.equals(siteId, studioConfiguration.getProperty(CONFIGURATION_GLOBAL_SYSTEM_SITE))) {
             content = configurationService.getGlobalConfigurationAsString(path);
         } else {
             content = configurationService.getConfigurationAsString(siteId, module, path, environment);
         }
+
+        if (content == null) {
```
jmendeza: Move the check and exception throwing to the service layer.

jmendeza (changes requested):

src/main/java/org/craftercms/studio/controller/rest/v2/ConfigurationController.java
phuongnq: Updated to throw exception for method `getConfigurationAsString`

jmendeza (approved):

PR #3159, opened 2023-01-05
Improving validation errors responses

jmendeza:
Improving validation errors responses
Groups and Logger Controllers validations
Removing unused methods
DependencyController update


src/main/java/org/craftercms/studio/model/rest/ValidationFieldError.java
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (C) 2007-2022 Crafter Software Corporation. All Rights Reserved.
```
sumerjabri: 2023

sumerjabri (changes requested):
Please update the copyright year for the changed files.

sumerjabri (approved):

PR #3165, opened 2023-01-11
Add group externally managed flag

phuongnq:
https://github.com/craftersoftware/craftercms/issues/483


src/main/java/org/craftercms/studio/impl/v2/service/security/GroupServiceImpl.java
```diff
@@ -160,6 +161,9 @@ public void deleteGroup(List<Long> groupIds)
                 throw new ServiceLayerException("Deleting the System Admin group is not allowed.");
             }
         }
+
+        checkExternallyManagedGroup(groupIds);
+
```
sumerjabri, line 166: Must also make sure we return to the UI the fact that a group is externally managed so it can disable the user's ability to make updates. Is that done already?

sumerjabri (changes requested):

jmendeza (approved):

src/main/java/org/craftercms/studio/impl/v2/service/security/GroupServiceImpl.java
phuongnq: @sumerjabri Yes, we are returning `externallyManaged` flag (`true` or `false`) to UI. See screenshot:

sumerjabri (approved):

PR #3171, opened 2023-01-16
Create API to check if site has initial publish

phuongnq:
https://github.com/craftercms/craftercms/issues/5847

```
GET /studio/api/2/publish/has_initial_publish?siteId=ed
{
    "response": {
        "code": 0,
        "message": "OK",
        "remedialAction": "",
        "documentationUrl": ""
    },
    "hasInitialPublish": true
}
```

### DIFF
--- a/src/main/java/org/craftercms/studio/controller/rest/v2/PublishController.java
+++ b/src/main/java/org/craftercms/studio/controller/rest/v2/PublishController.java
@@ -17,6 +17,7 @@
 package org.craftercms.studio.controller.rest.v2;
 
 import org.apache.commons.collections.CollectionUtils;
+import org.craftercms.commons.validation.annotations.param.EsapiValidatedParam;
 import org.craftercms.studio.api.v1.exception.ServiceLayerException;
 import org.craftercms.studio.api.v1.exception.SiteNotFoundException;
 import org.craftercms.studio.api.v1.exception.security.UserNotFoundException;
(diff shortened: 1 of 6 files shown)


src/main/java/org/craftercms/studio/controller/rest/v2/PublishController.java
```diff
@@ -215,6 +207,23 @@ public ResponseBody getAvailablePublishingTargets(@RequestParam(name = REQUEST_P
         return responseBody;
     }
 
+    @ValidateParams
```
jmendeza: @ValidateParams annotation has been removed. Please annotate the Controller class @Validated (from org.springframework.validation.annotation)

src/main/java/org/craftercms/studio/controller/rest/v2/PublishController.java
```diff
@@ -215,6 +207,23 @@ public ResponseBody getAvailablePublishingTargets(@RequestParam(name = REQUEST_P
...
+    @GetMapping(value = HAS_INITIAL_PUBLISH, produces = APPLICATION_JSON_VALUE)
+    public ResponseBody hasInitialPublish(@EsapiValidatedParam(type = SITE_ID) @RequestParam(name = REQUEST_PARAM_SITEID) String siteId)
+        throws SiteNotFoundException {
+        if (!siteService.exists(siteId)) {
```
jmendeza: Let's throw these exceptions from the service layer
Let's use
```siteService.checkSiteExists(siteId);```

jmendeza (changes requested):

src/main/java/org/craftercms/studio/controller/rest/v2/PublishController.java
phuongnq: Updated to use `@Validated` and `@Valid`

src/main/java/org/craftercms/studio/controller/rest/v2/PublishController.java
phuongnq: Moved check site exists to services layer. Thanks.

jmendeza (approved):

PR #3175, opened 2023-01-19
Refactoring proxy engine service internal

phuongnq:
https://github.com/craftersoftware/craftercms/issues/478

* Never copy the headers authorization, x-xsrf-token from the request to engine
* Modify Cookie header to remove xsrf-token, JSESSIONID (Crafter Studio's), refresh_token
* This service is currently being used only by the endpoint `/studio/engine/api/1/monitoring/log.json`. If the is no special reason, I don't think we need configurable headers from `proxy-config.xml`

Note: depends on the commons PR: https://github.com/craftercms/commons/pull/327

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/service/proxy/internal/ProxyServiceInternalImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/service/proxy/internal/ProxyServiceInternalImpl.java
@@ -17,6 +17,7 @@
 package org.craftercms.studio.impl.v2.service.proxy.internal;
 
 import org.apache.commons.lang3.StringUtils;
+import org.craftercms.commons.proxy.ProxyUtils;
 import org.craftercms.studio.api.v1.service.configuration.ServicesConfig;
 import org.craftercms.studio.api.v2.service.proxy.ProxyService;
 import org.craftercms.studio.api.v2.utils.StudioConfiguration;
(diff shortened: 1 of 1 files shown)


src/main/java/org/craftercms/studio/impl/v2/service/proxy/internal/ProxyServiceInternalImpl.java
```diff
@@ -69,49 +74,104 @@ public ResponseEntity<Object> getSiteLogEvents(final String body,
...
-                    resHeaders.addAll(key, value);
+        // rebuild cookie headers to remove ignored list of cookies
+        String proxyCookieHeader = "";
+        Cookie[] cookies = request.getCookies();
+        for (Cookie cookie : cookies) {
+            if (!IGNORE_REQUEST_COOKIES.contains(cookie.getName().toLowerCase())) {
+                if (StringUtils.isEmpty(proxyCookieHeader)) {
+                    proxyCookieHeader = format("%s=%s", cookie.getName(), cookie.getValue());
+                } else {
+                    proxyCookieHeader += format("; %s=%s", cookie.getName(), cookie.getValue());
```
jmendeza: There might be many cookies, let's use StringBuilder instead of `+=`

jmendeza (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/proxy/internal/ProxyServiceInternalImpl.java
phuongnq: Updated to use `StringBuilder`. Thanks!

phuongnq: This function has been moved to commons: https://github.com/craftercms/commons/pull/327

jmendeza (approved):

sumerjabri (approved):

PR #3184, opened 2023-01-25
Update engine sample header mappings config

phuongnq:
https://github.com/craftercms/craftercms/issues/5909

### DIFF
--- a/src/main/resources/crafter/studio/upgrade/pipelines.yaml
+++ b/src/main/resources/crafter/studio/upgrade/pipelines.yaml
@@ -484,6 +484,15 @@ pipelines:
             path: /configuration/global-permission-mappings-config.xml
             template: crafter/studio/upgrade/4.0.x/4.0.1.15/system/global-permission-mappings-config.xslt
             commitDetails: Remove cmis-related permissions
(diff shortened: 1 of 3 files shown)


sumerjabri (approved):

src/main/resources/crafter/studio/upgrade/pipelines.yaml
```diff
@@ -463,6 +463,14 @@ pipelines:
         operations:
           - type: dbScriptUpgrader
             filename: upgrade-4.0.1.13-to-4.0.1.14.sql
+      - currentVersion: 4.0.1.14
```
jmendeza: Please pull the latest and update the version number. There is already a v4.0.1.15

jmendeza (changes requested):

avasquez614 (approved):

src/main/resources/crafter/studio/upgrade/pipelines.yaml
phuongnq: Updated to new version `4.0.1.16`. Thanks.

jmendeza (approved):

PR #3189, opened 2023-02-01
Fix error 500 on request publish

jmendeza:
https://github.com/craftersoftware/craftercms/issues/470
https://github.com/craftersoftware/craftercms/issues/494
https://github.com/craftersoftware/craftercms/issues/493


src/main/java/org/craftercms/studio/impl/v2/service/workflow/WorkflowServiceImpl.java
```diff
@@ -242,8 +243,11 @@ private void createWorkflowEntries(String siteId, List<String> paths, String sub
             throws UserNotFoundException, ServiceLayerException {
         User userObj = userServiceInternal.getUserByIdOrUsername(-1, submittedBy);
         List<Workflow> workflowEntries = new LinkedList<>();
-        paths.forEach(path -> {
+        for (String path : paths) {
             Item it = itemServiceInternal.getItem(siteId, path);
+            if (it == null) {
+                throw new ContentNotFoundException(path, siteId, format("Failed to retrieve item at path '%s' in site '%s'", path, siteId));
```
sumerjabri, line 249: Will that fail the entire publishing queue?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/workflow/WorkflowServiceImpl.java
jmendeza: It will fail the current request. Notice this is request publish, so they are not in the queue just yet

sumerjabri (approved):

PR #3193, opened 2023-02-09
Correct socket event on writeContent after persisting content

phuongnq:
https://github.com/craftercms/craftercms/issues/5853

`writeContent` method is incorrectly sent ContentEvent right after it is written to the Git repo. We should only send notification when the item is persisted (write to DB)

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v1/service/content/ContentService.java
+++ b/src/main/java/org/craftercms/studio/api/v1/service/content/ContentService.java
@@ -136,6 +136,13 @@ public interface ContentService {
     boolean writeContent(String site, String path, InputStream content) throws ServiceLayerException;
 
     /**
+     * Notify when there is a content update
+     * @param site site name
+     * @param path path name
+     */
+    void notifyContentEvent(String site, String path);
+
+    /**
      * create a folder
      *
      * @param site - the project ID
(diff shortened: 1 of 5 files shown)


jmendeza (commented):
I see the calls to `notifyContentEvent()` happen right after `persistItemAfterWrite` or `persistItemAfterCreate`
It seems to me that we are giving the responsibility of triggering the event to different clients of the contentService/itemServiceInternal. 
Would it make sense to move `notifyContentEvent()` call to `itemServiceInternal` ?

jmendeza (changes requested):
Please take a look at my previous comment

phuongnq:
There is one concern for me that if we move them to `persistItemAfterWrite` and `persistItemAfterCreate` there are cases that the socket events are sending twice or more to an item change.

For example in `writeContentAsset`: https://github.com/craftercms/studio/blob/develop/src/main/java/org/craftercms/studio/impl/v1/content/pipeline/AssetDmContentProcessor.java#L118.
We first write the content and then update the content again in the following line:
* https://github.com/craftercms/studio/blob/develop/src/main/java/org/craftercms/studio/impl/v1/content/pipeline/AssetDmContentProcessor.java#L145 
* https://github.com/craftercms/studio/blob/develop/src/main/java/org/craftercms/studio/impl/v1/content/pipeline/AssetDmContentProcessor.java#L177

If we put both notification on `persistItemAfterWrite` and `persistItemAfterCreate` then we have a duplicated socket events.

Actually, in the past we had an issue that the socket events is duplicated: https://github.com/craftercms/craftercms/issues/5559 and solved it by moving the socket event to upper layers (public services). In general, I also think it is better to have socket events in more upper layers and near the public services as much as possible as mentioned here: https://github.com/craftercms/craftercms/issues/5559#issuecomment-1146721574

jmendeza (approved):

PR #3198, opened 2023-03-01
Prevent individual failures to stop publish_all operation

jmendeza:
https://github.com/craftercms/craftercms/issues/5949
Prevent individual failures to stop publish_all operation


src/main/java/org/craftercms/studio/api/v2/repository/RepositoryChanges.java
```diff
@@ -19,6 +19,7 @@
 
 import org.apache.commons.collections4.CollectionUtils;
 
+import java.util.HashSet;
```
sumerjabri: Is this the right data structure to use? Do we have concerns about the size of this set in memory? Should we consider TreeSet.

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/api/v2/repository/RepositoryChanges.java
jmendeza: As discussed, we are changing this to use ArrayList by default, since we the uniqueness should already be guaranteed by the updated/deleted path Sets

sumerjabri (approved):

PR #3201, opened 2023-03-07
Update publisher to item states are updated after successfully publishing each package

jmendeza:
https://github.com/craftercms/craftercms/issues/5949
Update publisher to item states are updated after successfully publishing each package


src/main/java/org/craftercms/studio/impl/v1/service/deployment/PublishingManagerImpl.java
```diff
@@ -90,6 +93,35 @@ public List<PublishRequest> getItemsReadyForDeployment(@ValidateStringParam Stri
         return publishRequestMapper.getItemsReadyForDeployment(params);
     }
 
+    /**
+     * Indicates if environment is the live/prod environment
+     * for the given site
+     * @param site the site id
+     * @param env the publishing environment
```
sumerjabri, line 100: please use `Publishing Target` instead of env

sumerjabri (approved):

PR #3206, opened 2023-03-09
Fix Dashboard APIs

phuongnq:
https://github.com/craftercms/craftercms/issues/5985

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -5547,10 +5547,10 @@ paths:
                     type: integer
                     format: int32
                     description: Number of records returned in this response
-                  publishingPackages:
+                  publishingItems:
                     type: array
                     items:
-                      $ref: '#/components/schemas/DashboardPublishingPackage'
+                      $ref: '#/components/schemas/DetailedItem'
         '400':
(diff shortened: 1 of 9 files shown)


src/main/api/2/studio-api-2.yaml
```diff
@@ -9057,6 +9057,41 @@ components:
         - size
         - publishingTarget
 
+    DashboardPublishRequestItem:
```
sumerjabri: Is it possible to return `DashboardPublishingPackage`? If so, let's do that.

sumerjabri: Please put the `submissioncomment` field as the `label` for now because `label` seems to be always `null`.

phuongnq: > Is it possible to return DashboardPublishingPackage? If so, let's do that.

=> Yes, it was the wrong design when I sent the items instead of packages to the response. Fixed to use 
`DashboardPublishingPackage`

> Please put the submissioncomment field as the label for now because label seems to be always null.

=> I added a field `comment` to put `submissioncomment` into. `label` seems to have different meaning, let me know if we want to remove the `label` from the response!

sumerjabri (approved):

PR #3211, opened 2023-03-20
Fix Dashboard APIs

phuongnq:
https://github.com/craftercms/craftercms/issues/5991
https://github.com/craftercms/craftercms/issues/5985

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/service/dashboard/DashboardServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/service/dashboard/DashboardServiceImpl.java
@@ -52,6 +52,7 @@ import org.craftercms.studio.model.search.SearchResult;
 import java.beans.ConstructorProperties;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import static co.elastic.clients.elasticsearch._types.SortOrder.Asc;
@@ -224,7 +225,7 @@ public class DashboardServiceImpl implements DashboardService {
     public ExpiringContentResult getContentExpiring(@ProtectedResourceId(SITE_ID_RESOURCE_ID) String siteId,
                                                     ZonedDateTime dateFrom, ZonedDateTime dateTo,
                                                     int offset, int limit)
-            throws AuthenticationException, ServiceLayerException {
+            throws AuthenticationException, ServiceLayerException, UserNotFoundException {
         siteService.checkSiteExists(siteId);
         SearchParams searchParams = new SearchParams();
         String query = getContentExpiringQuery()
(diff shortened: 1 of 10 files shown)


phuongnq:
@sumerjabri This PR includes https://github.com/craftercms/studio/pull/3206, no rollback for 3206 is required.

src/main/java/org/craftercms/studio/impl/v2/service/dashboard/DashboardServiceImpl.java
```diff
@@ -224,52 +225,76 @@ protected void prepareSearchParams(SearchParams searchParams, String query, Stri
...
-        return publishServiceInternal.getPublishingPackagesScheduled(siteId, publishingTarget, dateFrom, dateTo,
-                offset, limit);
+        var items =
+                publishServiceInternal.getPublishingItemsScheduled(siteId, publishingTarget, approver, dateFrom, dateTo, offset, limit);
+        if (items.isEmpty()) {
+            return emptyList();
+        }
+
+        var paths = items.stream().map(PublishRequest::getPath).collect(toList());
+        List<DetailedItem> result = new ArrayList<>();
```
jmendeza, line 291: Why is the `contentServiceInternal.getItemByPath` in a separate loop? Instead of including it as a map operation of the stream?

src/main/java/org/craftercms/studio/impl/v2/service/dashboard/DashboardServiceImpl.java
```diff
@@ -309,17 +334,23 @@ public List<DashboardPublishingPackage> getPublishingHistory(
...
-        // TODO: Try to do a single query
-        var publishingPackageDetails =
-                publishServiceInternal.getPublishingPackageDetails(siteId, publishingPackageId);
-        var paths = publishingPackageDetails.getItems().stream()
-                .map(PublishingPackageDetails.PublishingPackageItem::getPath)
+        var packageDetail = publishServiceInternal.getPublishingHistoryDetail(siteId, publishingPackageId, offset, limit);
+        var paths = packageDetail.stream()
+                .map(PublishRequest::getPath)
                 .collect(toList());
         if (isEmpty(paths)) {
```
jmendeza: Can this `isEmpty` check move up before the stream (check if package detail is empty)?

jmendeza (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/dashboard/DashboardServiceImpl.java
phuongnq: Fixed. Thank you!

src/main/java/org/craftercms/studio/impl/v2/service/dashboard/DashboardServiceImpl.java
phuongnq: This could depend on how we handle the exception. Since the function `contentServiceInternal.getItemByPath` throws some type of exceptions, if we use it in the map stream operation, we will need to do the handling such as:
```
try {
    List<DetailedItem> detailedItems =  items.stream().map(item -> {
        String path = item.getPath();
        try {
            return contentServiceInternal.getItemByPath(siteId, path, false);
        } catch (ServiceLayerException | UserNotFoundException e) {
            throw new RuntimeException(e);
        }
    }).collect(toList());

    return detailedItems;
} catch (RuntimeException e) {
    throw new ServiceLayerException(e);
}
```

I think this is a bit unclean but please let me know if you have any recommendations or thoughts.

jmendeza: Got it. You're right. Thank you

jmendeza (approved):

PR #3255, opened 2023-06-01
Global actions are checked against the system_admin group membership instead of permissions

phuongnq:
https://github.com/craftersoftware/craftercms/issues/551


sumerjabri (changes requested):
Please don't use roles nor groups to allow/deny action. Instead, see if we can use the permissions. For example, use `create-site` to allow the creation of a site. If the user has `create-site`, then the user can create a site, etc.

jmendeza:
> 

I believe we should update `create-site` to be consistent with the other permissions format: `create_site`. 
Also add (and enforce) a `delete_site` permission.

jmendeza (changes requested):
Permissions should be checked instead of groups membership

sumerjabri:
@phuongnq if we switch `create-site` to `create_site` to be consistent, we must make sure UM fixes that in existing installations.

sumerjabri:
CC @russdanner

src/main/resources/crafter/studio/upgrade/4.1.x/4.1.2/system/global-permission-mappings-config.xslt
```diff
@@ -0,0 +1,111 @@
...
+            </xsl:if>
+            <xsl:if test="not(permission = 'cancel_publish')">
+                <xsl:element name="permission">
+                    <xsl:text>cancel_publish</xsl:text>
+                </xsl:element>
+                <xsl:text>&#10;</xsl:text>
+            </xsl:if>
+            <xsl:if test="not(permission = 'Change Content Type')">
+                <xsl:element name="permission">
+                    <xsl:text>Change Content Type</xsl:text>
```
sumerjabri: Please switch this to `change_content_type` like you did `create-site`

src/main/webapp/repo-bootstrap/global/configuration/global-permission-mappings-config.xml
```diff
@@ -92,6 +92,16 @@
                 <permission>view_logs</permission>
                 <permission>view_log_levels</permission>
                 <permission>configure_log_levels</permission>
+                <permission>content_create</permission>
+                <permission>get_publishing_queue</permission>
+                <permission>cancel_publish</permission>
+                <permission>Change Content Type</permission>
```
sumerjabri: Let's fix this too.

sumerjabri (changes requested):

src/main/resources/crafter/studio/upgrade/4.1.x/4.1.2/system/global-permission-mappings-config.xslt
phuongnq: Updated. I also change the following permissions to match our convention:
```
S3 Read => s3_read
S3 Write => s3_write
```

src/main/webapp/repo-bootstrap/global/configuration/global-permission-mappings-config.xml
phuongnq: Updated. Thanks!

sumerjabri (approved):

PR #3265, opened 2023-06-08
Enforce Project Policy Configuration to all sections in the current sidebar

phuongnq:
https://github.com/craftercms/craftercms/issues/6104


src/main/webapp/repo-bootstrap/global/blueprints/1000_website_editorial/config/studio/site-policy-config.xml
```diff
@@ -16,7 +16,7 @@
   -->
 <site-policy>
     <statement>
-        <target-path-pattern>/static-assets/.*</target-path-pattern>
+        <target-path-pattern>/(?!scripts/.*\\.\\w+|templates/.*\\.\\w+|static-assets/app/.*\\.\\w+).*</target-path-pattern>
```
jmendeza: While working fine for `static-assets` and `pages`, this is not consistent for `scripts` and `templates`.

For a folder, a path like this will be turn lower case:
`/scripts/classes/somePackage`
However, it will be accepted if it is a file:
`/scripts/classes/somePackage/MyGroovyClassFile.groovy`

cc @sumerjabri

jmendeza (changes requested):

src/main/webapp/repo-bootstrap/global/blueprints/1000_website_editorial/config/studio/site-policy-config.xml
phuongnq: How about we enable the upper case for any folder in `/scripts`, `/templates` and `/static-assets/app`? This would make things more simple.

We could also update the regex to something like this `/(?!scripts/([a-z0-9_\\-]+/)*[\\w.]+\\.\\w+|templates/([a-z0-9_\\-]+/)*[\\w.]+\\.\\w+|static-assets/app/([a-z0-9_\\-]+/)*[\\w.]+\\.\\w+).*` to turn to lower case for the second test. But then it would transform from `/scripts/classes/somePackage/MyGroovyClassFile.groovy` to `/scripts/classes/somepackage/mygroovyclassfile.groovy` which I think we want to have `/scripts/classes/somepackage/MyGroovyClassFile.groovy`.

jmendeza: I think allowing mixed-case in folder names is a reasonable compromise between functionality and code/rules complexity.
Thoughts @sumerjabri ?

sumerjabri (approved):

jmendeza (approved):

PR #3267, opened 2023-06-12
Add dependent_items API v2

phuongnq:
https://github.com/craftercms/craftercms/issues/6061

From:
```
POST /api/1/services/api/1/dependency/get-dependant.json
```

To:

```
GET /api/2/dependency/dependent_items
```

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v2/service/dependency/DependencyService.java
+++ b/src/main/java/org/craftercms/studio/api/v2/service/dependency/DependencyService.java
@@ -64,7 +64,7 @@ public interface DependencyService {
      * @param paths list of paths to get dependent items for
      * @return list of paths dependent on given paths
      */
-    List<String> getDependentItems(String siteId, List<String> paths) throws SiteNotFoundException;
+    List<String> getDependentPaths(String siteId, List<String> paths) throws SiteNotFoundException;
 
     /**
      * Get all items that depend on this item
--- a/src/main/java/org/craftercms/studio/api/v2/service/dependency/DependencyService.java
+++ b/src/main/java/org/craftercms/studio/api/v2/service/dependency/DependencyService.java
@@ -18,6 +18,7 @@ package org.craftercms.studio.api.v2.service.dependency;
 
 import org.craftercms.studio.api.v1.exception.ServiceLayerException;
 import org.craftercms.studio.api.v1.exception.SiteNotFoundException;
+import org.craftercms.studio.model.rest.content.DependencyItem;
 
 import java.util.List;
 
@@ -64,4 +65,13 @@ public interface DependencyService {
      * @return list of paths dependent on given paths
      */
     List<String> getDependentItems(String siteId, List<String> paths) throws SiteNotFoundException;
+
+    /**
+     * Get all items that depend on this item
+     * @param siteId site identifier
+     * @param path path to get dependent items for
+     * @return list of items depending on given item path
+     * @throws ServiceLayerException Internal error, see exception details
+     */
+    List<DependencyItem> getDependentItems(String siteId, String path) throws ServiceLayerException;
 }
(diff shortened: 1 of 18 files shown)


src/main/java/org/craftercms/studio/controller/rest/v2/DependencyController.java
```diff
@@ -68,4 +69,17 @@ public ResponseBody getSoftDependencies(@RequestBody @Valid GetSoftDependenciesR
         responseBody.setResult(result);
         return responseBody;
     }
+
+    @GetMapping(DEPENDENT_ITEMS)
+    public ResponseBody getDependentItems(@NotEmpty @ValidSiteId @RequestParam String siteId,
```
jmendeza: What is the purpose of `ResponseBody`? It would simplify a bit to just return the `ResultOne`

src/main/java/org/craftercms/studio/impl/v2/service/dependency/DependencyServiceImpl.java
```diff
@@ -71,4 +76,12 @@ public List<String> getDependentItems(@ProtectedResourceId(SITE_ID_RESOURCE_ID)
         siteService.checkSiteExists(siteId);
         return dependencyServiceInternal.getDependentItems(siteId, paths);
     }
+
+    @Override
+    @HasPermission(type = DefaultPermission.class, action = PERMISSION_CONTENT_READ)
```
jmendeza, line 86: Don't we need to annotate the siteId and paths params with `@ProtectedResourceId`?

src/main/resources/crafter/studio/studio-services-context.xml
```diff
@@ -556,6 +556,7 @@
     <bean id="dependencyService" class="org.craftercms.studio.impl.v2.service.dependency.DependencyServiceImpl">
         <constructor-arg name="dependencyServiceInternal" ref="dependencyServiceInternal"/>
         <constructor-arg name="siteService" ref="cstudioSiteServiceSimple" />
+        <constructor-arg name="contentRepository" ref="gitContentRepositoryV2" />
```
jmendeza: We should use `contentRepository` here

jmendeza (changes requested):

src/main/java/org/craftercms/studio/controller/rest/v2/DependencyController.java
phuongnq: Fixed to use `ResultOne`. Thank you!

src/main/java/org/craftercms/studio/impl/v2/service/dependency/DependencyServiceImpl.java
phuongnq: We do need the `@ProtectedResourceId` here for `siteId` and `path` parameters. I fixed this. Thank you!

src/main/resources/crafter/studio/studio-services-context.xml
phuongnq: Fixed. Thank you!

jmendeza (approved):

src/main/java/org/craftercms/studio/api/v2/service/dependency/DependencyService.java
```diff
@@ -64,4 +65,13 @@ List<String> getHardDependencies(String site, List<String> paths)
      * @return list of paths dependent on given paths
      */
     List<String> getDependentItems(String siteId, List<String> paths) throws SiteNotFoundException;
```
jmendeza: Should we rename this method to prevent it from being confused with the new one?
`getDependentPaths` ?
Same for the the internal service interface

jmendeza (changes requested):

src/main/java/org/craftercms/studio/api/v2/service/dependency/DependencyService.java
phuongnq: I updated this method name to `getDependentPaths`. Thank you

phuongnq: I updated that method's name to `getDependentPaths`. Thank you.

jmendeza (approved):

PR #3272, opened 2023-06-15
Add /content/{siteId} API v2 to get content

phuongnq:
https://github.com/craftercms/craftercms/issues/6061

From
```
GET /api/1/services/api/1/content/get-content.json
```

To
```
GET /api/2/content/{siteId}?path=CONTENT_PATH&commitId=CONTENT_BY_COMMIT_ID
```


jmendeza (approved):

src/main/api/2/studio-api-2.yaml
```diff
@@ -4230,6 +4230,56 @@ paths:
...
+          required: true
+          schema:
+            type: string
+        - name: path
+          in: query
+          description: path of the content
+          required: true
+          schema:
+            type: string
+        - name: edit
```
jmendeza: What this really means is to lock the content so it is not edited by someone else.
Should we explicitly call this `lock` / `lockContent`?
@sumerjabri @phuongnq

jmendeza (changes requested):

src/main/api/2/studio-api-2.yaml
sumerjabri: We will create a small design doc to decide:
* API name
* Params and expectations

I think we should expect this one to work for any item (binary or text), and for locking to happen externally (lock, readItem, unlock).

More to come via design doc.

jmendeza: @phuongnq I have shared a doc with you. It is still in progress, but you can refer to the section about this API.

"Get content" is pretty much already implemented by "get_content_by_commit_id". Get content will rely on that functionality using HEAD as the commitId.
Required changes:
- Create new mapping in ContentController for a request like  "/content/{siteId}?path=abc&commitId=123". CommitId is optional, default to `HEAD`
- Keep "/get_content_by_commit_id" mapping, but mark it as deprecated.
- You might need to create a private method to perform the actual service invoking, since the mappings mix path and request parameters.
- Update studio-api-2.yaml accordingly. Include a note for deprecated URL.

Please let me know if you have any questions/suggestions/etc.


cc @sumerjabri

phuongnq: I updated the API as per the above discussion. Thank you!

src/main/api/2/studio-api-2.yaml
```diff
@@ -4230,11 +4230,61 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /content/{siteId}:
+    get:
+      tags:
+        - content
+      summary: Get content for a path
+      description: 'Required permission "content_read" and additional "content_write" if edit parameter is `true`'
```
jmendeza: edit parameter does not exist anymore

src/main/java/org/craftercms/studio/controller/rest/v2/ContentController.java
```diff
@@ -264,8 +265,27 @@ public ResponseBody itemUnlockByPath(@RequestBody @Valid UnlockItemByPathRequest
         return responseBody;
     }
 
+    @Valid
+    @GetMapping(PATH_PARAM_SITE)
+    public ResponseEntity<Resource> getContent(@ValidSiteId @PathVariable(value = REQUEST_PARAM_SITE) String siteId,
+                                          @ValidExistingContentPath @RequestParam(value = REQUEST_PARAM_PATH) String path,
+                                          @EsapiValidatedParam(type = ALPHANUMERIC) @RequestParam(value = REQUEST_PARAM_COMMIT_ID, required = false) String commitId)
+            throws ServiceLayerException, UserNotFoundException {
+        if (commitId == null) {
```
jmendeza: Let's use the @RequestParam `defaultValue` attribute for this

src/main/java/org/craftercms/studio/controller/rest/v2/ContentController.java
```diff
@@ -264,8 +265,27 @@ public ResponseBody itemUnlockByPath(@RequestBody @Valid UnlockItemByPathRequest
...
+            commitId = HEAD;
+        }
+        Resource resource = contentService.getContentByCommitId(siteId, path, commitId).orElseThrow();
```
jmendeza: Let's create a simple private method to prevent code duplication. `getContentByCommitId` and `getContent` methods are identical

jmendeza (changes requested):

src/main/api/2/studio-api-2.yaml
phuongnq: I updated this. Thank you!

src/main/java/org/craftercms/studio/controller/rest/v2/ContentController.java
phuongnq: I updated this. Thank you!

src/main/java/org/craftercms/studio/controller/rest/v2/ContentController.java
phuongnq: I updated this. Thank you!

jmendeza (approved):

src/main/api/2/studio-api-2.yaml
```diff
@@ -4230,11 +4230,61 @@ paths:
...
+          required: true
+          schema:
+            type: string
+        - name: path
+          in: query
+          description: path of the content
+          required: true
+          schema:
+            type: string
+        - name: commitId
```
jmendeza, line 4293: @phuongnq please also indicate the default value for commitId

src/main/api/2/studio-api-2.yaml
```diff
@@ -4230,11 +4230,61 @@ paths:
...
+          $ref: '#/components/responses/NotFound'
+        '500':
+          $ref: '#/components/responses/InternalServerError'
+
   /content/get_content_by_commit_id:
     get:
       tags:
         - content
-      summary: Get content for commit id
+      summary: Get content for commit id. (Deprecated. Use /content/{siteId} instead.)
```
jmendeza, line 4328: @sumerjabri please advise on the deprecation comment actual language

jmendeza (changes requested):

src/main/api/2/studio-api-2.yaml
phuongnq: I updated this. Thank you!

jmendeza: @phuongnq you can also use the `default` key in the parameter schema:
```
schema:
  type: string
  default: HEAD
```

phuongnq: I updated to use the schema default value. Thank you!

jmendeza (approved):

phuongnq:
Close in favor of https://github.com/craftercms/studio/pull/3355

PR #3273, opened 2023-06-17
Get cluster mode API doc

jmendeza:
https://github.com/craftersoftware/craftercms/issues/556
Get cluster mode API doc


src/main/api/2/studio-api-2.yaml
```diff
@@ -1817,7 +1817,37 @@ paths:
           $ref: '#/components/responses/NotFound'
         '500':
           $ref: '#/components/responses/InternalServerError'
-
+  /cluster/mode:
```
sumerjabri, line 1820: Isn't there a Management Token required here?

sumerjabri (changes requested):

src/main/api/2/studio-api-2.yaml
jmendeza: Updated. Thanks

sumerjabri (approved):

PR #3276, opened 2023-06-23
Added workflow submitter

jmendeza:
https://github.com/craftercms/craftercms/issues/6122
Added workflow submitter
Made creator, modifier and lockOwner Person objects instead of just the username


src/main/resources/org/craftercms/studio/api/v2/dal/ItemDAO.xml
```diff
@@ -49,6 +46,14 @@
         <result property="ignoredAsInt" column="ignored" />
         <result property="childrenCount" column="children_count" />
         <result property="previousPath" column="previous_path" />
+        <association property="modifier" column="modifier"
```
jmendeza, line 50: @sumerjabri please let me know your thoughts about these changes.
Each of these associations will translate to a User table query (although MyBatis is smart about it and will not request multiple times the same item).
The advantage is that it will reduce the number of joins in several queries and it will make it much easier to maintain.

jmendeza: I guess there might be some performance impact if having many items submitted by many different authors, otherwise it should remain pretty similar to what we have today.

sumerjabri: I think this is a major benefit. We may end up tuning the DB schema if this proves too expensive.

sumerjabri (approved):

jvega190:
This PR needs to be merged with https://github.com/craftercms/studio-ui/pull/3365 to avoid issues on the UI

PR #3277, opened 2023-06-26
[6138] Headless blog bp doesn't persist updated content via XB update

jvega190:
https://github.com/craftercms/craftercms/issues/6138


jvega190:
@rart

rart (approved):

PR #3288, opened 2023-07-18
Require site ready

jmendeza:
https://github.com/craftersoftware/craftercms/issues/549
Site lifecycle events
Require Site Ready
InvalidSiteStateException

This PR depends on
https://github.com/craftercms/commons/pull/349


src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
```diff
@@ -2109,6 +2109,8 @@ public boolean revertContentItem(@ValidateStringParam String site,
             if (isEmpty(commitId)) {
                 return false;
             }
+            String username = securityService.getCurrentUser();
```
sumerjabri: Is this related to this issue/PR?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
jmendeza: @sumerjabri no, it is not. I will exclude from this one

sumerjabri (approved):

PR #3296, opened 2023-07-31
Optimize Upgrade Manager configuration loading

jmendeza:
https://github.com/craftercms/craftercms/issues/6160
Updated the UpgradeManager implementation to use an UpgradeConfigurationProvider instead of loading the file directly


jmendeza:
Please notice this PR and https://github.com/craftercms/commons/pull/350 are interdependent

sumerjabri (approved):

PR #3301, opened 2023-08-15
Content item history V2

jmendeza:
https://github.com/craftercms/craftercms/issues/6153
Content item history V2


src/main/api/2/studio-api-2.yaml
```diff
@@ -9216,6 +9261,33 @@ components:
...
+        oldPath:
+          type: string
+          description: the old path of the item (will be the same as path if the item has not been renamed)
+        author:
+          type: string
+          description: the author of the commit
+        comment:
+          type: string
+          description: the comment of the commit
+        reversible:
```
sumerjabri: revertable

sumerjabri (changes requested):
Please replace reversible with revertable since it's more accurate.

Also, do we not need unit tests for this API?

sumerjabri (approved):

PR #3305, opened 2023-08-18
Forgot password asynchronous execution

jmendeza:
https://github.com/craftersoftware/craftercms/issues/583
Forgot password asynchronous execution


sumerjabri (changes requested):
Let's use a separate threadpool

sumerjabri (approved):

PR #3346, opened 2023-10-24
Use configuration type for config xml files

phuongnq:
https://github.com/craftercms/craftercms/issues/4502

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v2/security/ContentItemPossibleActionsConstants.java
+++ b/src/main/java/org/craftercms/studio/api/v2/security/ContentItemPossibleActionsConstants.java
@@ -24,7 +24,6 @@ import static org.craftercms.studio.api.v1.constant.StudioConstants.CONTENT_TYPE
 import static org.craftercms.studio.api.v1.constant.StudioConstants.CONTENT_TYPE_CONFIG_FOLDER;
 import static org.craftercms.studio.api.v1.constant.StudioConstants.CONTENT_TYPE_CONTENT_TYPE;
 import static org.craftercms.studio.api.v1.constant.StudioConstants.CONTENT_TYPE_DOCUMENT;
-import static org.craftercms.studio.api.v1.constant.StudioConstants.CONTENT_TYPE_FILE;
 import static org.craftercms.studio.api.v1.constant.StudioConstants.CONTENT_TYPE_FOLDER;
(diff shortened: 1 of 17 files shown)


jmendeza (approved):

sumerjabri (approved):

sumerjabri:
@jmendeza note the removal of the FILE constant. I didn't look deeply at this yet.

src/main/java/org/craftercms/studio/api/v2/security/ContentItemPossibleActionsConstants.java
```diff
@@ -24,7 +24,6 @@
 import static org.craftercms.studio.api.v1.constant.StudioConstants.CONTENT_TYPE_CONFIG_FOLDER;
 import static org.craftercms.studio.api.v1.constant.StudioConstants.CONTENT_TYPE_CONTENT_TYPE;
 import static org.craftercms.studio.api.v1.constant.StudioConstants.CONTENT_TYPE_DOCUMENT;
-import static org.craftercms.studio.api.v1.constant.StudioConstants.CONTENT_TYPE_FILE;
```
sumerjabri, line 27: Is this now not used? Is there a different catch-all?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/api/v2/security/ContentItemPossibleActionsConstants.java
phuongnq: @sumerjabri In the past I think we changed the type UNKNOWN to FILE: https://github.com/craftercms/studio/commit/ec91fa1ae7a603cff3548e6fcf0fd97a7450baa7

If there is no other type, a path will have `system_type` is FILE.

In this PR, FILE type is replaced with CONFIGURATION type for any config files, the remaining FILE type include the following if you create a new site from Editorial BP:
```
/.crafter/**
/.gitignore
/craftercms-plugin.yaml
/sources/**
```

By default FILE type will have no possible action at all:
```
case CONTENT_TYPE_UNKNOWN:
            default:
                toRet = 0L;
                break;
```

Let me know if we want to add some possible actions for the default FILE type.

jmendeza (approved):

sumerjabri (approved):

PR #3352, opened 2023-10-30
Remove Git commits from the database where possible

jmendeza:
https://github.com/craftercms/craftercms/issues/6126
Remove git commit ids from site table. Remove gitlog. Update syncFromRepo process


sumerjabri (changes requested):
We will wait for the commit ID's removal ID from the publish_request table after we merge the publisher enhancements into 4.2.

sumerjabri (approved):

PR #3377, opened 2023-11-21
Engine preview mode updates

jmendeza:
https://github.com/craftersoftware/craftercms/issues/504
Engine preview mode updates

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -5339,6 +5339,28 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /security/preview/switch:
+    post:
+      tags:
+        - security
+      summary: Configures preview to use a different site. Site is expected to be set in `crafterSite` cookie
+      operationId: switchPreviewSite
+      description: Required permission "LOGGED_IN"
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  response:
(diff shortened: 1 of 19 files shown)


src/main/api/2/studio-api-2.yaml
```diff
@@ -5339,6 +5339,28 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /security/preview/switch:
```
sumerjabri, line 5342: Why not /site/switch?

sumerjabri: Also, what happens if I call this API to switch to a site I'm not allowed? What would be the HTTP response code?

sumerjabri (changes requested):

src/main/api/2/studio-api-2.yaml
jmendeza: I added it to Security controller because it already had the `AccessTokenService` dependency. It is also a security concern.

If this API is called with a `crafterSite` the user has no access to (or empty/absent), it will remove the `crafterPreview` cookie.
Response code would be 200.

sumerjabri: Should the response code 404 for a site that doesn't exist indicating that the site is not found (not found for this user)?

jmendeza: Updated.
404 when user does not have access or site does not exist

sumerjabri (approved):

PR #3389, opened 2023-12-01
Send policy modified message reason

phuongnq:
https://github.com/craftercms/craftercms/issues/4730

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/service/policy/validators/PathPolicyValidator.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/service/policy/validators/PathPolicyValidator.java
@@ -21,6 +21,8 @@ import org.slf4j.LoggerFactory;
 import org.craftercms.studio.impl.v2.service.policy.PolicyValidator;
 import org.craftercms.studio.model.policy.Action;
 import org.craftercms.studio.model.policy.ValidationResult;
+
+import static java.lang.String.format;
 import static org.apache.commons.lang3.StringUtils.*;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/craftercms/studio/impl/v2/service/policy/validators/PathPolicyValidator.java
```diff
@@ -58,18 +60,27 @@ private void validatePermitted(HierarchicalConfiguration<?> config, Action actio
         var sourceRegex = config.getString(CONFIG_KEY_SOURCE_REGEX);
         if (!item.matches(sourceRegex)) {
             String modifiedItem = null;
+            StringBuilder reasonBuilder = new StringBuilder();
             var targetRegex = config.getString(CONFIG_KEY_TARGET_REGEX);
             if (targetRegex != null) {
                 modifiedItem = item.replaceAll(sourceRegex, targetRegex);
-
+                if (!modifiedItem.equalsIgnoreCase(item)) {
+                    addPolicyReason(reasonBuilder, format("Path is replaced from source regex '%s' to target regex '%s'.", sourceRegex, targetRegex));
```
sumerjabri: Let's change this to show the user the before and after. Something like "Path '%s' was transformed to '%s' per the project file name policy."

sumerjabri: The above can be used for all cases. No need for lowercase/uppercase.

src/main/java/org/craftercms/studio/impl/v2/service/policy/validators/PathPolicyValidator.java
```diff
@@ -58,18 +60,27 @@ private void validatePermitted(HierarchicalConfiguration<?> config, Action actio
...
+                }
                 var caseTransform = config.getString(CONFIG_KEY_CASE_TRANSFORM);
                 if (isNotEmpty(caseTransform)) {
                     switch (caseTransform.toLowerCase()) {
                         case "uppercase":
-                            modifiedItem = modifiedItem.toUpperCase();
+                            if (modifiedItem.chars().anyMatch(c -> Character.isLowerCase(c))) {
+                                addPolicyReason(reasonBuilder, "Path is transformed to uppercase.");
```
sumerjabri: Maybe we use the one reason as I mention above.

src/main/java/org/craftercms/studio/impl/v2/service/policy/validators/PathPolicyValidator.java
```diff
@@ -58,18 +60,27 @@ private void validatePermitted(HierarchicalConfiguration<?> config, Action actio
...
+                                modifiedItem = modifiedItem.toUpperCase();
+                            }
                             break;
                         case "lowercase":
-                            modifiedItem = modifiedItem.toLowerCase();
+                            if (modifiedItem.chars().anyMatch(c -> Character.isUpperCase(c))) {
+                                addPolicyReason(reasonBuilder, "Path is transformed to lowercase.");
```
sumerjabri: Maybe we use the one reason as I mention above.

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/policy/validators/PathPolicyValidator.java
phuongnq: I updated the reason message.

src/main/java/org/craftercms/studio/impl/v2/service/policy/validators/PathPolicyValidator.java
phuongnq: I updated the reason message.

src/main/java/org/craftercms/studio/impl/v2/service/policy/validators/PathPolicyValidator.java
phuongnq: I updated the reason message.

sumerjabri (approved):

PR #3390, opened 2023-12-05
Get Content Types API

jmendeza:
https://github.com/craftercms/craftercms/issues/6306
Get Content Types API


src/main/java/org/craftercms/studio/controller/rest/v2/ConfigurationController.java
```diff
@@ -78,13 +82,28 @@ public Result clearCache(@ValidSiteId @RequestParam String siteId) throws SiteNo
         return result;
     }
 
+    @GetMapping("/content-types/{siteId}/all")
```
sumerjabri: Let's change this to match the pattern we finally agree uponl

src/main/java/org/craftercms/studio/model/contentType/ContentTypeConfigFiles.java
```diff
@@ -0,0 +1,60 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Arrays;
+import java.util.Iterator;
+
+/**
+ * Represents the xml configuration files (config.xml, form-definition.xml) pair for a content type.
+ */
+public class ContentTypeConfigFiles implements Iterable<String> {
```
sumerjabri: We don't need the `config`, just the form definition.

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/controller/rest/v2/ConfigurationController.java
```diff
@@ -78,13 +82,28 @@ public Result clearCache(@ValidSiteId @RequestParam String siteId) throws SiteNo
...
+    public ResponseBody getAllContentTypes(@ValidSiteId @PathVariable("siteId") String siteId) throws ServiceLayerException {
+        ResponseBody responseBody = new ResponseBody();
+        long startTime = System.currentTimeMillis();
```
sumerjabri: Consider making this if trace is enabled.

sumerjabri (approved):

PR #3398, opened 2023-12-21
Content API v1 updates

jmendeza:
https://github.com/craftersoftware/craftercms/issues/675
Content API v1 updates


sumerjabri (changes requested):
Please also port/move shallowContentExists

sumerjabri (approved):

PR #3405, opened 2023-12-26
GetChildrenByPaths API

jmendeza:
https://github.com/craftersoftware/craftercms/issues/682
GetChildrenByPaths API

### DIFF
--- a/src/main/api/2/studio-api-2.yaml
+++ b/src/main/api/2/studio-api-2.yaml
@@ -3860,6 +3860,86 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /content/children_by_paths:
+    post:
+      tags:
+        - content
+      summary: Get list of children for given item paths
+      description: 'Required permission "get_children" and "site member"'
+      operationId: getChildrenByPaths
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                siteId:
+                  description: Site ID
+                  type: string
+                localeCode:
+                  description: |-
+                    filter children by locale code, or show source items (not translation) if translation to
+                    specified locale code does not exist
+                  type: string
+                paths:
+                  description: item paths to get children for
+                  type: array
+                  items:
+                    $ref: '#/components/schemas/ChildrenByPathRequest'
+              required:
+                - siteId
+                - paths
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  response:
+                    $ref: '#/components/schemas/ApiResponse'
+                  items:
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        path:
+                          type: string
+                        levelDescriptor:
+                          $ref: '#/components/schemas/SandboxItem'
+                        children:
+                          type: array
+                          items:
+                            $ref: '#/components/schemas/SandboxItem'
+                        total:
+                          type: integer
+                          format: int32
+                          description: Total records in the system
+                        offset:
+                          type: integer
+                          format: int32
+                          description: Offset of first record in this response
+                        limit:
+                          type: integer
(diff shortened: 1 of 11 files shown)


sumerjabri (approved):

src/main/api/2/studio-api-2.yaml
```diff
@@ -3860,6 +3860,86 @@ paths:
...
+              type: object
+              properties:
+                siteId:
+                  description: Site ID
+                  type: string
+                localeCode:
+                  description: |-
+                    filter children by locale code, or show source items (not translation) if translation to
+                    specified locale code does not exist
+                  type: string
```
rart: If we are going to refresh multiple widgets at once, we could require it per path. Per design, each widget has its own locale selector. It is not set globally but per widget.

Originally, the localeCode is part of the individual requests. Is there a performance reason to exact it from the individual request? Why change it?

src/main/api/2/studio-api-2.yaml
```diff
@@ -3860,6 +3860,86 @@ paths:
...
+                          format: int32
+                          description: Total records in the system
+                        offset:
+                          type: integer
+                          format: int32
+                          description: Offset of first record in this response
+                        limit:
+                          type: integer
+                          format: int32
+                          description: Number of records returned in this response
```
rart: We send a single call to `sandbox_items_by_path` to fetch all the parents at once, and 1 call per parent to get its children to `children_by_path`. We could save one more request if this new api would return the SandboxItem (parent) for that path too, albeit, I do believe there are cases where we only need to refresh the parent or only need to refresh the children so we call only the `item_by_path` or `children_by_path` alone.

src/main/api/2/studio-api-2.yaml
```diff
@@ -3860,6 +3860,86 @@ paths:
...
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                siteId:
+                  description: Site ID
+                  type: string
+                localeCode:
```
sumerjabri: Let's move the locale code inside the request array to allow for it to be different if required.

src/main/api/2/studio-api-2.yaml
```diff
@@ -3765,10 +3765,11 @@ paths:
 
   /content/children_by_path:
     post:
+      deprecated: true
       tags:
         - content
       summary: Get list of children for given item path
-      description: 'Required permission "get_children" and "site member"'
+      description: 'Required permission "get_children" and "site member".<br/> This operation is deprecated, use /content/{siteId}/children instead.'
```
sumerjabri, line 3772: Add `deprecated: true` instead of the comment.

sumerjabri (changes requested):

src/main/api/2/studio-api-2.yaml
jmendeza: `deprecated: true` is already there. Added the comment to indicate the alternative API endpoint

sumerjabri (approved):

jmendeza:
@rart @sumerjabri 
Please notice that this new API includes results that are currently retrieved from `sandbox_items_by_path`, which takes a `preferContent` parameter. I only found calls with that set to true.
This new API sets that internally to `true` when retrieving the sandbox items. Please let me know if there is a case where we would need `preferContent` to be false and therefore add another param.

sumerjabri (approved):

src/main/api/2/studio-api-2.yaml
```diff
@@ -3860,6 +3861,80 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /content/{siteId}/children:
+    post:
+      tags:
+        - content
+      summary: Get list of children for given item paths
```
rart:
```suggestion
      summary: Get the list of children for each of the requested parent paths
```

rart (approved):

PR #3430, opened 2024-01-09
Use RequireSiteExists annotation to check site

phuongnq:

* Use `RequireSiteExists` annotation instead of `siteService.checkSiteExists(site);`
* When a method has `RequireSiteReady`, there is no need to further check site exists since site ready has also check if the site first exists.

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/service/content/ContentServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/service/content/ContentServiceImpl.java
@@ -99,7 +99,7 @@ public class ContentServiceImpl implements ContentService, ApplicationContextAwa
     @Override
     @RequireSiteExists
     @HasPermission(type = DefaultPermission.class, action = PERMISSION_CONTENT_READ)
-    public boolean contentExists(@ProtectedResourceId(SITE_ID_RESOURCE_ID) @SiteId String siteId,
+    public boolean contentExists(@SiteId String siteId,
                                  @ProtectedResourceId(PATH_RESOURCE_ID) String path) throws SiteNotFoundException {
         return contentServiceInternal.contentExists(siteId, path);
     }
--- a/src/main/java/org/craftercms/studio/impl/v2/service/content/ContentServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/service/content/ContentServiceImpl.java
@@ -34,6 +34,7 @@ import org.craftercms.studio.api.v1.service.GeneralLockService;
 import org.craftercms.studio.api.v1.service.deployment.DeploymentException;
 import org.craftercms.studio.api.v1.service.deployment.DeploymentService;
 import org.craftercms.studio.api.v1.service.site.SiteService;
+import org.craftercms.studio.api.v2.annotation.RequireSiteExists;
 import org.craftercms.studio.api.v2.annotation.RequireSiteReady;
 import org.craftercms.studio.api.v2.annotation.SiteId;
 import org.craftercms.studio.api.v2.dal.AuditLog;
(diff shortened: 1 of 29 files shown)


src/main/java/org/craftercms/studio/impl/v2/service/content/ContentServiceImpl.java
```diff
@@ -96,23 +97,23 @@ public class ContentServiceImpl implements ContentService, ApplicationContextAwa
     private org.craftercms.studio.api.v1.service.content.ContentService contentServiceV1;
 
     @Override
+    @RequireSiteExists
     @HasPermission(type = DefaultPermission.class, action = PERMISSION_CONTENT_READ)
-    public boolean contentExists(@ProtectedResourceId(SITE_ID_RESOURCE_ID) String siteId,
+    public boolean contentExists(@ProtectedResourceId(SITE_ID_RESOURCE_ID) @SiteId String siteId,
```
jmendeza: `SiteId` itself is annotated with `@ProtectedResourceId`. Please validate just `@SiteId` would be enough

src/main/java/org/craftercms/studio/impl/v2/service/dependency/DependencyServiceImpl.java
```diff
@@ -79,14 +75,13 @@ public List<String> getDependentPaths(@SiteId String siteId,
...
     public List<String> getHardDependencies(@SiteId String site,
                                             @ProtectedResourceId(PATH_LIST_RESOURCE_ID) List<String> paths) throws ServiceLayerException {
-        siteService.checkSiteExists(site);
         return dependencyServiceInternal.getHardDependencies(site, paths);
     }
 
+    @RequireSiteExists
     @HasPermission(type = DefaultPermission.class, action = PERMISSION_CONTENT_READ)
-    public List<DependencyItem> getDependentItems(@ProtectedResourceId(SITE_ID_RESOURCE_ID) String siteId,
+    public List<DependencyItem> getDependentItems(@SiteId @ProtectedResourceId(SITE_ID_RESOURCE_ID) String siteId,
```
jmendeza: @SiteId itself is annotated with @ProtectedResourceId

src/main/java/org/craftercms/studio/impl/v2/service/site/SitesServiceImpl.java
```diff
@@ -85,14 +86,13 @@ public void updateSite(@SiteId String siteId, String name, String description)
...
-        checkSiteExists(siteId);
         sitesServiceInternal.updateSite(siteId, name, description);
     }
 
     @Override
+    @RequireSiteExists
     @HasPermission(type = DefaultPermission.class, action = PERMISSION_DELETE_SITE)
-    public void deleteSite(@ProtectedResourceId(SITE_ID_RESOURCE_ID) String siteId) throws ServiceLayerException {
-        checkSiteExists(siteId);
+    public void deleteSite(@SiteId @ProtectedResourceId(SITE_ID_RESOURCE_ID) String siteId) throws ServiceLayerException {
```
jmendeza: SiteId itself is annotated with @ProtectedResourceId. Please validate just @SiteId would be enough

jmendeza (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/content/ContentServiceImpl.java
phuongnq: Updated to have `@SiteId` only

src/main/java/org/craftercms/studio/impl/v2/service/dependency/DependencyServiceImpl.java
phuongnq: Updated to have `@SiteId` only

src/main/java/org/craftercms/studio/impl/v2/service/site/SitesServiceImpl.java
phuongnq: Updated to have `@SiteId` only

jmendeza (approved):

sumerjabri (approved):

PR #3436, opened 2024-01-11
Recursive soft dependency calculation

jmendeza:
Recursive soft dependency calculation
https://github.com/craftercms/craftercms/issues/6189

### DIFF
--- a/src/main/resources/crafter/studio/studio-config.yaml
+++ b/src/main/resources/crafter/studio/studio-config.yaml
@@ -589,7 +589,8 @@ studio.db.retryingOperation.maxSleep: 1000
 studio.db.scriptRunner.bufferSize: 10000
 # Batch size for bulk operations
 studio.db.bulkOperations.batchSize: 1000
-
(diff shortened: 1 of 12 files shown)


src/main/resources/crafter/studio/studio-config.yaml
```diff
@@ -589,7 +589,8 @@ studio.db.retryingOperation.maxSleep: 1000
 studio.db.scriptRunner.bufferSize: 10000
 # Batch size for bulk operations
 studio.db.bulkOperations.batchSize: 1000
-
+# DB max_recursive_iterations value. This property should be set to a value between 0 and 20 (hard limit)
+studio.db.maxRecursiveIterations: 0
```
sumerjabri: What does 0 do?

sumerjabri (changes requested):

src/main/resources/crafter/studio/studio-config.yaml
jmendeza: It should be 10. Fixed.

sumerjabri (approved):

PR #3440, opened 2024-01-15
Fix "studio-api.yaml" file version

alhambrav:
Fix "studio-api.yaml" file version https://github.com/craftercms/craftercms/issues/6423


sumerjabri (approved):
Does this mean we

PR #3455, opened 2024-01-22
Add RequireContentExists annotation

phuongnq:

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v2/annotation/RequireContentExists.java
+++ b/src/main/java/org/craftercms/studio/api/v2/annotation/RequireContentExists.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright (C) 2007-2024 Crafter Software Corporation. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as published by
+ * the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.craftercms.studio.api.v2.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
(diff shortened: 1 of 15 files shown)


sumerjabri (approved):

src/main/java/org/craftercms/studio/api/v2/annotation/RequireContentExists.java
```diff
@@ -0,0 +1,27 @@
...
+package org.craftercms.studio.api.v2.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.METHOD, ElementType.TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface RequireContentExists {
```
jmendeza, line 27: Will it work If we annotate this with `@RequireSiteExists`? So we don't need both everywhere?

src/main/java/org/craftercms/studio/api/v2/annotation/RequireContentExistsAnnotationHandler.java
```diff
@@ -0,0 +1,81 @@
...
+            "within(@RequireContentExists *) || " +
+            "within(@(@RequireContentExists *) *) || " +
+            "@annotation(RequireContentExists) || " +
+            "execution(@(@RequireContentExists *) * *(..))")
+    public Object requireContentExists(ProceedingJoinPoint pjp) throws Throwable {
+        Method method = AopUtils.getActualMethod(pjp);
+        String siteId = StudioAnnotationUtils.getAnnotationValue(pjp, method, SiteId.class, String.class);
+        String path = StudioAnnotationUtils.getAnnotationValue(pjp, method, ContentPath.class, String.class);
+
+        if (StringUtils.isNotEmpty(siteId) && StringUtils.isNotEmpty(path)) {
```
jmendeza: Can we handle the case where siteId is an empty string meaning the global repo should be used?

jmendeza (changes requested):

src/main/java/org/craftercms/studio/api/v2/annotation/RequireContentExists.java
phuongnq: I updated it so that we only need `@RequireContentExists` to check both site exists and content exists.

src/main/java/org/craftercms/studio/api/v2/annotation/RequireContentExistsAnnotationHandler.java
phuongnq: I updated to only check `path` is empty. If `siteId` is empty, we will check the content in the Global repo.

jmendeza (approved):

PR #3461, opened 2024-01-23
Add annotation to trace execution time

phuongnq:

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v2/annotation/LogExecutionTime.java
+++ b/src/main/java/org/craftercms/studio/api/v2/annotation/LogExecutionTime.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright (C) 2007-2024 Crafter Software Corporation. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as published by
+ * the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.craftercms.studio.api.v2.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.METHOD, ElementType.TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface LogExecutionTime {
+}
--- a/src/main/java/org/craftercms/studio/api/v2/annotation/LogExecutionTime.java
+++ b/src/main/java/org/craftercms/studio/api/v2/annotation/LogExecutionTime.java
@@ -21,6 +21,11 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+/**
+ * Annotation to trace execution time of methods.
+ * This annotation only works with public method with external calls as a limitation of Spring AOP.
+ * If the tracing method is an internal call, consider using {@link org.craftercms.studio.impl.v2.utils.TimeUtils} instead
+ */
 @Target({ElementType.METHOD, ElementType.TYPE})
 @Retention(RetentionPolicy.RUNTIME)
 public @interface LogExecutionTime {
--- a/src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
@@ -56,7 +56,6 @@ import org.craftercms.studio.api.v1.service.security.SecurityService;
 import org.craftercms.studio.api.v1.service.site.SiteService;
 import org.craftercms.studio.api.v1.to.RemoteRepositoryInfoTO;
 import org.craftercms.studio.api.v1.to.SiteBlueprintTO;
-import org.craftercms.studio.api.v2.annotation.LogExecutionTime;
 import org.craftercms.studio.api.v2.annotation.RequireSiteExists;
 import org.craftercms.studio.api.v2.annotation.SiteId;
 import org.craftercms.studio.api.v2.dal.*;
@@ -343,9 +342,9 @@ public class SiteServiceImpl implements SiteService, ApplicationContextAware {
         auditServiceInternal.insertAuditLog(auditLog);
     }
 
-    @LogExecutionTime
     private void processCreatedFiles(String siteId, Map<String, String> createdFiles, String creator,
                                      ZonedDateTime now) {
+        long startProcessCreatedFilesMark = logger.isDebugEnabled() ? System.currentTimeMillis() : 0L;
         SiteFeed siteFeed;
         try {
             siteFeed = getSite(siteId);
@@ -433,6 +432,10 @@ public class SiteServiceImpl implements SiteService, ApplicationContextAware {
 
             studioDBScriptRunner.execute(createdFileScriptPath.toFile());
             studioDBScriptRunner.execute(updateParentIdScriptPath.toFile());
+            if (logger.isDebugEnabled()) {
+                logger.debug("ProcessCreatedFiles finished in '{}' milliseconds",
+                        (System.currentTimeMillis() - startProcessCreatedFilesMark));
+            }
         } catch (IOException | ServiceLayerException e) {
             logger.error("Failed to create the database script file for processingCreatedFiles in site '{}'", siteId, e);
         } finally {
--- a/src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
@@ -78,6 +78,7 @@ import org.craftercms.studio.api.v2.utils.StudioConfiguration;
 import org.craftercms.studio.api.v2.utils.StudioUtils;
 import org.craftercms.studio.impl.v2.utils.DateUtils;
 import org.craftercms.studio.impl.v2.utils.DependencyUtils;
+import org.craftercms.studio.impl.v2.utils.TimeUtils;
 import org.craftercms.studio.model.blobstore.BlobStoreDetails;
 import org.craftercms.studio.model.site.SiteDetails;
 import org.dom4j.Document;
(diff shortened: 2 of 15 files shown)


src/main/java/org/craftercms/studio/impl/v2/event/EventBroadcaster.java
```diff
@@ -52,17 +53,10 @@ public void publishGlobalEvent(final GlobalBroadcastEvent event) {
         publishEvent(event, DESTINATION_ROOT);
     }
 
+    @LogExecutionTime
     private void publishEvent(final BroadcastEvent event, final String destination) {
```
jmendeza, line 60: Does it work with private methods?

src/main/java/org/craftercms/studio/impl/v2/repository/GitContentRepository.java
```diff
@@ -353,11 +354,11 @@ public String getRepoFirstCommitId(final String site) {
         return toReturn;
     }
 
+    @LogExecutionTime
```
jmendeza: Does it work with private methods?

src/main/java/org/craftercms/studio/impl/v2/service/configuration/ConfigurationServiceImpl.java
```diff
@@ -352,11 +347,8 @@ public String getGlobalConfigurationAsString(@ProtectedResourceId(PATH_RESOURCE_
         return content;
     }
 
+    @LogExecutionTime
```
jmendeza: Does it work with private methods?

jmendeza (changes requested):

src/main/java/org/craftercms/studio/impl/v2/event/EventBroadcaster.java
phuongnq: @jmendeza It does not work with private methods. Do we have any concerns if I update these methods to public?

jmendeza: @phuongnq I believe `protected` should do, let's try that first

sumerjabri:
@phuongnq it seems that we have conflicts.

src/main/java/org/craftercms/studio/impl/v2/event/EventBroadcaster.java
phuongnq: @jmendeza It seems we are using Spring AOP Proxies and the annotation does not trigger on the internal method call regardless it is public/protected/private method: https://docs.spring.io/spring-framework/reference/core/aop/proxying.html#aop-understanding-aop-proxies
For this particular method, I have moved the `@LogExecutionTime` to the upper public method.
For other places, I have reverted the change and I'm currently looking for other solutions. Please let me know if you have any recommendations! 

We can for example have a lambda to wrap the function like this:
```
TimeUtils.logExecutionTime(() -> method(param1, param2));
```
But I found it is not very clean.

Another solution such as using load-time weaving with AspectJ: https://docs.spring.io/spring-framework/reference/core/aop/using-aspectj.html#aop-aj-ltw but this will make some structure change required.

jmendeza: I see. Let's try load-time weaving. We should measure differences in startup time and memory footprint though.
Can we do compile-time weaving?

jmendeza: I see. Let's try load-time weaving. We should measure differences in startup time and memory footprint though.
Can we do compile-time weaving?

jmendeza (changes requested):

src/main/java/org/craftercms/studio/impl/v2/event/EventBroadcaster.java
phuongnq: @jmendeza I have succeeded in using load time weaver but there are some concerns:
* It adds some overhead while starting up Studio:
```
[ParallelWebappClassLoader@3918c187] warning javax.* types are not being woven because the weaver option '-Xset:weaveJavaxPackages=true' has not been specified
05-Feb-2024 21:44:11.777 INFO [main] org.apache.jasper.servlet.TldScanner.scanJars At least one JAR was scanned for TLDs yet contained no TLDs. Enable debug logging for this logger for a complete list of JARs that were scanned but no TLDs were found in them. Skipping unneeded JARs during scanning can improve startup time and JSP compilation time.
[INFO] 2024-02-05T21:45:43,260 [main] [util.Version] | HV000001: Hibernate Validator 6.2.5.Final 
```
=> In this log, it took 1.5 minutes to complete the weaving (it could take up to 3 minutes in some cases).
* We need to add the `javaagent` to `CATALINA_OPTS` to make sure the aspect is weaving with AspectJ methods. An example of the configure such as:
```
export CATALINA_OPTS=${CATALINA_OPTS:="-server -Xss1024K -Xms1G -Xmx4G -Dlog4j2.formatMsgNoLookups=true -javaagent:/home/pnguyen/.m2/repository/org/aspectj/aspectjweaver/1.9.21/aspectjweaver-1.9.21.jar"}
```
=> It's not great if we need to carry a `jar` file to add to the startup options.

* We need to register which class to be weaved in the file `META-INF/aop.xml` like this:
```
<!DOCTYPE aspectj PUBLIC "-//AspectJ//DTD//EN" "https://www.eclipse.org/aspectj/dtd/aspectj.dtd">
<aspectj>

    <weaver options="-showWeaveInfo -verbose">
        <!-- only weave classes in our application-specific packages -->
        <include within="org.craftercms.studio.api.v2.annotation.*"/>
        <include within="org.craftercms.studio.impl.v2.event.EventBroadcaster"/>
    </weaver>

    <aspects>
        <!-- weave in just this aspect -->
        <aspect name="org.craftercms.studio.api.v2.annotation.LogExecutionTimeAnnotationHandler"/>
    </aspects>

</aspectj>
```
Adding a class such as `EventBroadcaster` will cause a list here and it is not clean in terms of maintainability and may be hard to debug.

phuongnq: Regarding compile-time weaving, I'm not optimistic with our project yet. We will need to add the following plugin to `pom.xml`:

```
<plugin>
				<groupId>org.codehaus.mojo</groupId>
				<artifactId>aspectj-maven-plugin</artifactId>
				<version>1.15.0</version>
				<executions>
					<execution>
						<goals>
							<goal>compile</goal>       <!-- use this goal to weave all your main classes -->
							<goal>test-compile</goal>  <!-- use this goal to weave all your test classes -->
						</goals>
					</execution>
				</executions>
				<configuration>
					<source>16</source>
					<target>16</target>
					<complianceLevel>17.0</complianceLevel>
					<showWeaveInfo>true</showWeaveInfo>
					<encoding>UTF-8</encoding>
				</configuration>
			</plugin>
```
But while run `mvn clean install`, I got some error:
```
[INFO] Showing AJC message detail for messages of types: [error, warning, fail]
[WARNING] bad version number found in /home/pnguyen/.m2/repository/org/aspectj/aspectjrt/1.9.21/aspectjrt-1.9.21.jar expected 1.9.7 found 1.9.21
	<unknown source file>:<no line information>

[ERROR] around on staticinitialization of interface 'org.craftercms.studio.api.v2.annotation.RequireSiteReady' not supported (compiler limitation)
	/home/pnguyen/craftercms/next/craftercms/src/studio/src/main/java/org/craftercms/studio/api/v2/annotation/RequireSiteStateAnnotationHandler.java:58

[ERROR] around on staticinitialization of interface 'org.craftercms.studio.api.v2.annotation.RequireSiteReady' not supported (compiler limitation)
	/home/pnguyen/craftercms/next/craftercms/src/studio/src/main/java/org/craftercms/studio/api/v2/annotation/RequireSiteReady.java:35
public @interface RequireSiteReady {
                  ^^^^^^^^^^^^^^^

[ERROR] around on staticinitialization of interface 'org.craftercms.studio.api.v2.annotation.RequireContentExists' not supported (compiler limitation)
	/home/pnguyen/craftercms/next/craftercms/src/studio/src/main/java/org/craftercms/studio/api/v2/annotation/RequireSiteExistsAnnotationHandler.java:60

[ERROR] around on staticinitialization of interface 'org.craftercms.studio.api.v2.annotation.RequireContentExists' not supported (compiler limitation)
	/home/pnguyen/craftercms/next/craftercms/src/studio/src/main/java/org/craftercms/studio/api/v2/annotation/RequireContentExists.java:27
public @interface RequireContentExists {
                  ^^^^^^^^^^^^^^^^^^^

[INFO] Join point 'method-execution(org.craftercms.studio.model.rest.ResponseBody org.craftercms.studio.controller.rest.v2.ConfigurationController.getConfiguration(java.lang.String, java.lang.String, java.lang.String, java.lang.String))' in Type 'org.craftercms.studio.controller.rest.v2.ConfigurationController' (ConfigurationController.java:88) advised by around advice from 'org.craftercms.studio.api.v2.annotation.LogExecutionTimeAnnotationHandler' (LogExecutionTimeAnnotationHandler.java:50)
[INFO] Join point 'method-execution(org.craftercms.studio.model.rest.ResponseBody org.craftercms.studio.controller.rest.v2.ModelController.getModelDefinitions(java.lang.String))' in Type 'org.craftercms.studio.controller.rest.v2.ModelController' (ModelController.java:52) advised by around advice from 'org.craftercms.studio.api.v2.annotation.LogExecutionTimeAnnotationHandler' (LogExecutionTimeAnnotationHandler.java:50)
[ERROR] The package org.xml.sax is accessible from more than one module: <unnamed>, java.xml
	/home/pnguyen/craftercms/next/craftercms/src/studio/src/main/java/org/craftercms/studio/impl/v1/util/ContentUtils.java:25
import org.xml.sax.SAXException;
       ^^^^^^^^^^
```
This could due to the plugin is not fully supporting Java 17 or/and needing some more configuration where we do the compile/where to not.
Note that I also tried the other plugin as some suggestion in the above plugin github but the result is the same: https://github.com/dev-aspectj/aspectj-maven-plugin

sumerjabri:
@phuongnq are you stuck on this one?

phuongnq:
@sumerjabri I was able to run the LTW but have some concerns. Please see my above comment in the first review thread

src/main/java/org/craftercms/studio/impl/v2/event/EventBroadcaster.java
phuongnq: cc @sumerjabri

sumerjabri:
@jmendeza please see Phil's comment

src/main/java/org/craftercms/studio/impl/v2/event/EventBroadcaster.java
jmendeza: @phuongnq I'm curious why are `target` and `source` set to `16` in the plugin configuration

But anyway, that path seems more complex than expected and it would be adding more moving pieces, i.e. future maintenance. Not a great trade-off in my opinion.
I believe having some helper method like the suggested is an acceptable compromise for the cases where the annotation won't work:
```
TimeUtils.logExecutionTime(() -> method(param1, param2));
```
@sumerjabri

phuongnq: Updated to have additional `TimeUtils.logExecutionTime()` method.
This lambda also has some limitations in that we should not run it against functions with exceptions in the signature because we then have to try/catch the RunTimeException inside the lambda and it could mean loss of the original exception instance.

For new implementing methods now on, I think we should design them so that it is `public` and external call to reuse the annotation. That would make the code clean.

FYI
> I'm curious why are target and source set to 16 in the plugin configuration

This is due to this commit: https://github.com/mojohaus/aspectj-maven-plugin/pull/155/files#diff-a03bee29a28d152f1791ed2a8b35c38152dee87a5d429604007ccd748c417544. For some reason, they use target and source 16 to support java 17.

sumerjabri: @jmendeza, we need to document somewhere for us and future developers how to weave this into the methods to trace execution time cleanly, as Phil indicated.

src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
```diff
@@ -353,110 +345,108 @@ private void insertCreateSiteAuditLog(String siteId, String siteName, String blu
 
     private void processCreatedFiles(String siteId, Map<String, String> createdFiles, String creator,
```
jmendeza, line 349: Let's extract to another method so we don't have a huge lambda. Either that or add the `logExecutionTime` call to the caller

jmendeza (changes requested):

jmendeza (approved):

src/main/java/org/craftercms/studio/api/v2/annotation/LogExecutionTime.java
```diff
@@ -0,0 +1,27 @@
...
+package org.craftercms.studio.api.v2.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.METHOD, ElementType.TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface LogExecutionTime {
```
jmendeza, line 31: Let's add some comments about what this annotation does and what the requirements of the annotated method are so it works properly

src/main/java/org/craftercms/studio/impl/v2/utils/TimeUtils.java
```diff
@@ -0,0 +1,42 @@
...
+     * High order function to inject execution time if the logger is enabled trace
+     * @param supplier method to calculate the execution time
+     * @param methodLogger logger of the method
+     * @param message a message to flag the execution method
+     * @return
+     * @param <T>
+     */
+    public static <T> T logExecutionTime(Supplier<T> supplier, Logger methodLogger, String message) {
+        long startTime = 0;
+        if (methodLogger.isTraceEnabled()) {
```
jmendeza, line 34: Why is this trace and the annotation is debug?

src/main/java/org/craftercms/studio/impl/v2/utils/TimeUtils.java
```diff
@@ -0,0 +1,42 @@
...
+public class TimeUtils {
+    /**
+     * High order function to inject execution time if the logger is enabled trace
+     * @param supplier method to calculate the execution time
+     * @param methodLogger logger of the method
+     * @param message a message to flag the execution method
+     * @return
+     * @param <T>
+     */
+    public static <T> T logExecutionTime(Supplier<T> supplier, Logger methodLogger, String message) {
```
jmendeza, line 32: We should note here that the preferred approach is to use the annotation when possible

jmendeza (changes requested):

jmendeza (approved):

jmendeza (approved):

PR #3485, opened 2024-02-17
Handle TooManyNestedClausesSearchException

jmendeza:
https://github.com/craftersoftware/craftercms/issues/786
Handle TooManyNestedClausesSearchException


src/main/java/org/craftercms/studio/impl/v2/service/search/internal/ClausesLimitAwareSearchService.java
```diff
@@ -0,0 +1,60 @@
...
+
+    protected final SearchService actualSearchService;
+
+    @ConstructorProperties({"actualSearchService"})
+    public ClausesLimitAwareSearchService(final SearchService actualSearchService) {
+        this.actualSearchService = actualSearchService;
+    }
+
+    @Override
+    public SearchResult search(final String siteId, final SearchParams params, final int initialMaxExpansions) throws ServiceLayerException {
```
sumerjabri, line 48: We should add a log indicating that we hit this limit at level WARN so admins know we're hitting a limit, and then retry. If we completely fail, then we log ERROR.

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/search/internal/ClausesLimitAwareSearchService.java
```diff
@@ -0,0 +1,69 @@
...
+                return actualSearchService.search(siteId, params, maxExpansions);
+            } catch (Exception e) {
+                TooManyNestedClausesSearchException tooManyClausesException = getThrowableOfType(e, TooManyNestedClausesSearchException.class);
+                if (tooManyClausesException == null) {
+                    throw e;
+                }
+                lastException = tooManyClausesException;
+                maxExpansions = maxExpansions / 2;
+                if (maxExpansions >= 1) {
+                    logger.warn("Search query for site '{}' contains too many nested clauses, " +
```
sumerjabri: Please log the max_expansions parameter so we have better insight.

sumerjabri (changes requested):

sumerjabri (approved):

PR #3490, opened 2024-02-21
New publish APIs

jmendeza:
New publish APIs


src/main/api/studio-api.yaml
```diff
@@ -10356,27 +10552,42 @@ components:
...
+      type: object
+      properties:
+        siteId:
+          type: string
+          description: site ID
+        publishingTarget:
+          type: string
+          description: the publishing target (staging or live)
+          example: staging
         items:
```
sumerjabri: `paths`

src/main/api/studio-api.yaml
```diff
@@ -10356,27 +10552,42 @@ components:
...
+                description: the path of the item to publish (mutually exclusive with commitId)
+              includeChildren:
+                type: boolean
+                description: include children of the given path (applicable only when publishing a path)
+                default: false
+              includeSoftDeps:
+                type: boolean
+                description: include soft dependencies (also includes soft dependencies of children)
+                default: false
+              commitId:
```
sumerjabri: move to `commitIds` array

src/main/api/studio-api.yaml
```diff
@@ -6545,6 +6545,133 @@ paths:
...
+                  type: string
+                  description: the publishing target (staging or live)
+                schedule:
+                  type: string
+                  format: date-time
+                  description: schedule the publish
+                comment:
+                    type: string
+                    description: optional comment
+                items:
```
sumerjabri: `paths`

src/main/api/studio-api.yaml
```diff
@@ -6545,6 +6545,133 @@ paths:
...
+                        description: the path of the item to publish (mutually exclusive with commitId)
+                      includeChildren:
+                        type: boolean
+                        description: include children of the given path (applicable only when publishing a path)
+                        default: false
+                      includeSoftDeps:
+                        type: boolean
+                        description: include soft dependencies (also includes soft dependencies of children)
+                        default: false
+                      commitId:
```
sumerjabri: move to `commitId` array

src/main/api/studio-api.yaml
```diff
@@ -6545,6 +6545,133 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /api/2/publish/calculate:
```
sumerjabri: `/publish/dependencies`

sumerjabri (changes requested):

src/main/api/studio-api.yaml
jmendeza: Done

src/main/api/studio-api.yaml
jmendeza: Done

src/main/api/studio-api.yaml
jmendeza: Done

src/main/api/studio-api.yaml
jmendeza: Done

src/main/api/studio-api.yaml
jmendeza: Done

sumerjabri (changes requested):
Let's discuss:

I really don't like the mixing of commitId and path based publishing/dep calc, etc. I think we need two root APIs, one around commits and the other paths. The reasons are:

1. Documentation is easier to understand and more meaningful
2. Test automation is better when the params are one and clear (including security and penetration testing)
3. There is no use-case where the user is mixing both, unless via the UI and the UI can unpack a commit ID first and then submit publish by path
4. API aesthetics

LMK your thoughts before I push forward with this.
The change becomes:

- Calc for commitId returns all the files, hard deps, soft deps
- Publish by commitId is the commit ID (maybe plural)
- Publish takes only paths (which may have come by expanding the commitId and submitting files)

jmendeza:
Added a `requestApproval` param to publish API. If false, user needs publish permission

jmendeza:
Moved to PR for new feature branch https://github.com/craftercms/studio/pull/3544

PR #3497, opened 2024-03-11
[4.1] Update search API request param

alhambrav:
Update search API request param https://github.com/craftersoftware/craftercms/issues/786


jmendeza (changes requested):
We ended up deciding against this parameter in the API. The param will be set (starting on 50) in the backed and search will be retried as needed until we get to `1` or get a success response from OS.
cc @sumerjabri

alhambrav:
Closing PR, re: response above from @jmendeza

PR #3501, opened 2024-03-16
Support variables in configuration files. Duplicate site serverless support

jmendeza:
https://github.com/craftercms/craftercms/issues/6630
https://github.com/craftercms/craftercms/issues/6624
Support variables in configuration files
Force RW blobstores when delivery serverless mode is on
Studio duplicate target requests updates (send create-target parameters)


src/main/webapp/repo-bootstrap/global/configuration/samples/sample-blob-stores-config.xml
```diff
@@ -62,36 +62,28 @@
...
             <mapping>
                 <publishingTarget>preview</publishingTarget>
-                <storeTarget>my-authoring-bucket</storeTarget>
-                <prefix>sandbox</prefix>
-            </mapping>
-            <mapping>
-                <publishingTarget>staging</publishingTarget>
-                <storeTarget>my-authoring-bucket</storeTarget>
-                <prefix>staging</prefix>
+                <storeTarget>${env:PREVIEW_BUCKET_NAME_PATTERN}</storeTarget>
```
avasquez614, line 74: Since we now have a staging env variable, should we add back the staging target?

avasquez614 (changes requested):

src/main/webapp/repo-bootstrap/global/configuration/samples/sample-blob-stores-config.xml
jmendeza: Added back...

sumerjabri (approved):

sumerjabri (approved):

src/main/resources/crafter/studio/studio-config.yaml
```diff
@@ -105,6 +105,8 @@ studio.repo.git.cli.process.destroy.waitForTimeout: 60
 studio.blob.config.module: studio
 # Location of the blob store configuration file in the site
 studio.blob.config.path: blob-stores-config.xml
+# Location of the default blob store configuration file if the site does not have one on site create
+studio.blob.default.config.path: classpath:crafter/studio/serverless/default-blob-stores-config.xml
```
avasquez614, line 109: Can I override this file by putting it under `CRAFTER_INSTALL/shared/classes/crafter/studio/serverless/default-blob-stores-config.xml`? Or I have to change the property to the exact path in the file system? (like `file:/opt/crafter/blabla...`)

avasquez614 (changes requested):

src/main/resources/crafter/studio/studio-config.yaml
jmendeza: You have to change the property. You can use a classpath location though. Something like:
```
studio.blob.default.config.path: classpath:crafter/studio/serverless/override-blob-stores-config.xml
```

avasquez614 (approved):

PR #3502, opened 2024-03-25
upgrade to spring 6

phuongnq:
https://github.com/craftercms/craftercms/issues/5977


src/main/java/org/craftercms/studio/impl/v2/service/aws/mediaconvert/AwsMediaConvertServiceImpl.java
```diff
@@ -100,33 +100,19 @@ public class AwsMediaConvertServiceImpl extends AbstractAwsService<MediaConvertP
...
-    @Required
-    public void setDashExtension(final String dashExtension) {
         this.dashExtension = dashExtension;
+        this.smoothExtension = smoothExtension;
     }
 
-    @Required
-    public void setSmoothExtension(final String smoothExtension) {
-        this.smoothExtension = smoothExtension;
+    public void setPartSize(final int partSize) {
```
sumerjabri, line 114: Could you let me know if you are keeping this outside the constructor because it is used that way?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/aws/mediaconvert/AwsMediaConvertServiceImpl.java
phuongnq: @sumerjabri Same reason for the Engine. We only migrate setter with the annotation `@Required`. For this particular setter, we don't use it at all: https://github.com/craftercms/studio/blob/develop/src/main/resources/crafter/studio/studio-services-context.xml#L1273 in the bean initialization

sumerjabri (approved):

sumerjabri (approved):

sumerjabri (approved):

avasquez614 (approved):

PR #3503, opened 2024-03-25
[6565] Fix updating filters breaking XB

jvega190:
https://github.com/craftercms/craftercms/issues/6565


jvega190:
@rart I realized that bootstrap's glyphicons fonts were missing from the bp, so the icons (for example in filters) were not showing. Added them back

rart (approved):

rart (approved):

sumerjabri (approved):

PR #3510, opened 2024-04-11
Write content before triggering the rename lifecycle event

phuongnq:
https://github.com/craftercms/craftercms/issues/6674


src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
```diff
@@ -1063,15 +1063,16 @@ public String moveContent(@ValidateStringParam @SiteId String site,
                     }
                 }
 
+                // This write is performed after processing the commitIds so we don't miss any commit
```
jmendeza: That comment is no longer relevant in 4.2

jmendeza (approved):

sumerjabri (approved):

PR #3522, opened 2024-04-24
ConfigurableMax annotation. write_configuration max length configurable

jmendeza:
ConfigurableMax annotation. Make write_configuration max length configurable
https://github.com/craftercms/craftercms/issues/6710

### DIFF
--- a/src/main/java/org/craftercms/studio/controller/rest/v2/ConfigurationController.java
+++ b/src/main/java/org/craftercms/studio/controller/rest/v2/ConfigurationController.java
@@ -111,7 +111,7 @@ public class ConfigurationController {
 
     @Valid
     @PostMapping("/write_configuration")
-    public ResponseBody writeConfiguration(@Valid @RequestBody WriteConfigurationRequest wcRequest)
+    public ResponseBody writeConfiguration(@Validated @RequestBody WriteConfigurationRequest wcRequest)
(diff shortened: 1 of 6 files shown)


src/main/java/org/craftercms/studio/controller/rest/v2/ConfigurationController.java
```diff
@@ -111,7 +111,7 @@ public ResponseBody getConfiguration(@ValidSiteId @RequestParam(name = "siteId",
 
     @Valid
     @PostMapping("/write_configuration")
-    public ResponseBody writeConfiguration(@Valid @RequestBody WriteConfigurationRequest wcRequest)
+    public ResponseBody writeConfiguration(@Validated @RequestBody WriteConfigurationRequest wcRequest)
```
jmendeza, line 114: Notice that @Validated is meant to work as an extension of @Valid. For some reason, custom validator is being triggered twice per request when using @Validated

sumerjabri: Do you mean @valid causes double invocation, so you used @validated?

jmendeza: Yes

avasquez614 (approved):

sumerjabri (approved):

PR #3545, opened 2024-05-10
Upgrade to AWS SDK v2

phuongnq:
https://github.com/craftercms/craftercms/issues/6617

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/repository/blob/s3/StudioAwsS3BlobStore.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/repository/blob/s3/StudioAwsS3BlobStore.java
@@ -67,9 +67,12 @@ public class StudioAwsS3BlobStore extends AwsS3BlobStore implements StudioBlobSt
 
     protected boolean readOnly;
 
-    @ConstructorProperties({"servicesConfig"})
-    public StudioAwsS3BlobStore(final ServicesConfig servicesConfig) {
+    private final ThreadPoolTaskExecutor taskExecutor;
+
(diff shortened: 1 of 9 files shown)


sumerjabri (approved):

sumerjabri (approved):

src/main/java/org/craftercms/studio/impl/v2/repository/blob/s3/StudioAwsS3BlobStore.java
```diff
@@ -69,7 +70,7 @@ public class StudioAwsS3BlobStore extends AwsS3BlobStore implements StudioBlobSt
     private final ThreadPoolTaskExecutor taskExecutor;
 
     @ConstructorProperties({"servicesConfig", "taskExecutor"})
-    public StudioAwsS3BlobStore(final ServicesConfig servicesConfig, final ThreadPoolTaskExecutor taskExecutor) {
+    public StudioAwsS3BlobStore(final ServicesConfig servicesConfig, ThreadPoolTaskExecutor taskExecutor) {
```
jmendeza: Any reason why removing the `final` ?

jmendeza (approved):

src/main/java/org/craftercms/studio/impl/v2/repository/blob/s3/StudioAwsS3BlobStore.java
phuongnq: fixed

avasquez614 (approved):

PR #3556, opened 2024-06-04
Audit session timeout event

phuongnq:
https://github.com/craftercms/craftercms/issues/6552

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/security/listener/AbstractAuditListener.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/security/listener/AbstractAuditListener.java
@@ -55,9 +55,10 @@ public abstract class AbstractAuditListener {
     }
 
     protected void recordAuthenticationEvent(String operation, AbstractAuthenticationEvent event, String message) {
+        String systemSite = studioConfiguration.getProperty(CONFIGURATION_GLOBAL_SYSTEM_SITE);
         try {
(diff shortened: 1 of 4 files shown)


src/main/java/org/craftercms/studio/impl/v2/security/listener/AbstractAuditListener.java
```diff
@@ -74,4 +75,32 @@ protected void recordAuthenticationEvent(String operation, AbstractAuthenticatio
...
+            auditLog.setPrimaryTargetId(username);
+            auditLog.setPrimaryTargetType(TARGET_TYPE_USER);
+            auditLog.setPrimaryTargetValue(username);
+            auditServiceInternal.insertAuditLog(auditLog);
+
+            if (isNotEmpty(message)) {
+                logger.info(message, name);
+            }
+        } catch (SiteNotFoundException e) {
+            // This never happens
```
jmendeza: Nonetheless, let's log the error as behavior might silently change in the future

jmendeza (changes requested):

src/main/java/org/craftercms/studio/impl/v2/security/listener/AbstractAuditListener.java
phuongnq: Updated

sumerjabri (approved):

PR #3561, opened 2024-06-11
Use RequireSiteExists annotation to check site

phuongnq:

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/service/dashboard/DashboardServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/service/dashboard/DashboardServiceImpl.java
@@ -220,7 +220,7 @@ public class DashboardServiceImpl implements DashboardService {
 
     @Override
     @RequireSiteExists
-    public ExpiringContentResult getContentExpired(String siteId, int offset, int limit)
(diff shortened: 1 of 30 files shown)


src/main/java/org/craftercms/studio/impl/v2/service/dashboard/DashboardServiceImpl.java
```diff
@@ -224,9 +219,9 @@ public ExpiringContentResult getContentExpiring(@SiteId String siteId,
     }
 
     @Override
+    @RequireSiteExists
     public ExpiringContentResult getContentExpired(String siteId, int offset, int limit)
```
jmendeza: Missing `@SiteId` annotation in the parameter?

jmendeza (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/dashboard/DashboardServiceImpl.java
phuongnq: Fixed. Thanks!

jmendeza (approved):

sumerjabri (approved):

PR #3565, opened 2024-06-18
Replace and remove DASHBOARD permissions

phuongnq:
https://github.com/craftersoftware/craftercms/issues/847


src/main/resources/crafter/studio/upgrade/pipelines.yaml
```diff
@@ -1096,6 +1096,12 @@ configurations:
             - type: xsltFileUpgrader
               template: crafter/studio/upgrade/4.1.x/permission-mappings/permission-mappings-config-v4.1.2.xslt
               commitDetails: Change 'Change Content Type' permission to 'change_content_type'
+        - currentVersion: 4.1.2
+          nextVersion: 4.2.0
```
jmendeza: Please make this 4.2.0.0. (we would then continue with 4.2.0.1 and so on)

jmendeza (changes requested):

src/main/java/org/craftercms/studio/permissions/StudioPermissionsConstants.java
```diff
@@ -45,6 +45,7 @@ public final class StudioPermissionsConstants {
     public static final String PERMISSION_LIST_REMOTES = "list_remotes";
     public static final String PERMISSION_PUBLISH = "publish";
     public static final String PERMISSION_PUBLISH_STATUS = "publish_status";
+    public static final String PERMISSION_PUBLISH_BY_COMMITS = "publish_by_commits";
```
jmendeza, line 48: Please note in the api spec that this permission is now required

src/main/resources/crafter/studio/upgrade/4.2.x/permission-mappings/permission-mappings-config-v4.2.0.xslt
```diff
@@ -0,0 +1,59 @@
...
+        </xsl:copy>
+    </xsl:template>
+
+    <!-- Remove <rule> elements with regex="~DASHBOARD~" -->
+    <xsl:template match="rule[@regex='~DASHBOARD~']"/>
+
+    <!-- Match role elements to process their children -->
+    <xsl:template match="role">
+        <xsl:variable name="hasDashboardRule" select="rule[@regex='~DASHBOARD~']/allowed-permissions/permission = 'publish'" />
+        <xsl:variable name="hasPublishByCommits" select="rule[@regex='.*']/allowed-permissions/permission = 'publish_by_commits'" />
```
jmendeza: Isn't `publish_by_commits` just being added in this PR?

jmendeza (changes requested):

src/main/resources/crafter/studio/upgrade/pipelines.yaml
phuongnq: Updated to use 4.2.0.0

src/main/java/org/craftercms/studio/permissions/StudioPermissionsConstants.java
phuongnq: I updated this.

src/main/resources/crafter/studio/upgrade/4.2.x/permission-mappings/permission-mappings-config-v4.2.0.xslt
phuongnq: Right. But I think it won't cause any issue with this defending logic. We don't want to have a duplicated `publish_by_commits` in the permission list in any case.

jmendeza (approved):

sumerjabri (approved):

PR #3569, opened 2024-07-17
Remove sync-from-repo dependency on audit table: new processed_commits table

jmendeza:
https://github.com/craftercms/craftercms/issues/6856
Remove sync-from-repo dependency on audit table: new processed_commits table

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/GeneralLockServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/GeneralLockServiceImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2007-2023 Crafter Software Corporation. All Rights Reserved.
+ * Copyright (C) 2007-2024 Crafter Software Corporation. All Rights Reserved.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 3 as published by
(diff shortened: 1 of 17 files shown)


src/main/java/org/craftercms/studio/impl/v1/service/GeneralLockServiceImpl.java
```diff
@@ -61,6 +63,20 @@ public void lock(@ValidateStringParam String objectId) {
         }
     }
 
+    @Override
+    public void lock(final String... lockKeys) {
+        logger.debug("Thread '{}' will attempt to lock multiple objects '{}'", Thread.currentThread().getName(), lockKeys);
+        Arrays.stream(lockKeys).sorted().forEach(this::lock);
```
sumerjabri, line 69: Imagine a case where:
* Caller 1 asks for ABCD locks
* Caller 2 asks for DEB locks simultaneously
* Deadlock
How do you avoid deadlocks?

Perhaps making this method synchronized is best.

Also, what happens if one of the locks fails or is locked? Do you wait forever?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v1/service/GeneralLockServiceImpl.java
jmendeza: We are sorting the `lockKeys` to try to prevent the issue with ABCD - DEB.
It will wait until the resource is unlocked. Problem with making this method `synchronized` is that it will prevent other threads from acquiring any locks, even if unrelated.

sumerjabri (approved):

PR #3581, opened 2024-07-31
Remove repo locks from ContentService. Add sync-from-repo lock to Publish submission

jmendeza:
Remove repo locks from ContentService.
Add sync-from-repo lock to Publish submission
https://github.com/craftercms/craftercms/issues/6856

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/sync/RepositoryWatcherImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/sync/RepositoryWatcherImpl.java
@@ -116,7 +116,6 @@ public class RepositoryWatcherImpl implements RepositoryWatcher, ApplicationEven
     /**
      * Creates a {@link QueuedEvent} for the given site and adds it to the queue.
      *
-     * @param siteId
      */
     private void queueRepoEvent(String siteId) {
         QueuedEvent event = queuedEvents.get(siteId);
@@ -147,6 +146,7 @@ public class RepositoryWatcherImpl implements RepositoryWatcher, ApplicationEven
             if (!queuedEvent.additionalEvents().get()) {
                 break;
             }
+            queuedEvent.additionalEvents().set(false);
             resetCount++;
         }
         queuedEvents.remove(queuedEvent.siteId());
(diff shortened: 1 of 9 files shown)


src/main/java/org/craftercms/studio/impl/v2/sync/RepositoryWatcherImpl.java
```diff
@@ -147,6 +146,7 @@ private void processRepoEvent(final QueuedEvent queuedEvent) {
             if (!queuedEvent.additionalEvents().get()) {
                 break;
             }
+            queuedEvent.additionalEvents().set(false);
```
sumerjabri, line 149: What's this for?

src/main/resources/org/craftercms/studio/api/v2/dal/PublishRequestDAO.xml
```diff
@@ -158,18 +158,6 @@
         LIMIT #{offset}, #{limit}
     </select>
 
-    <select id="getScheduledItems" resultMap="PublishRequestMap" parameterType="java.util.Map">
```
sumerjabri, line 161: Why are we removing this?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v2/sync/RepositoryWatcherImpl.java
jmendeza: The `additionalEvents` flag is set to true if an event is received while another one is already in the queue for the site.
The idea is to throttle the processing of events when there is constant activity. So the processing will be delayed either until no new events happen during a period (`getEventHandlingDelayMillis()`) or when the max reset count is reached.
We need to clear the flag each time we wait. So processing can proceed if no new event was received in the last sleep period.

src/main/resources/org/craftercms/studio/api/v2/dal/PublishRequestDAO.xml
jmendeza: This code is not used anymore since the `get-scheduled-items` API was retired
For reference https://github.com/craftercms/studio/pull/3348/files#diff-943cf4769cda17d2d9e595c881f9413dfcbb70c968d59f371ac0cfd947fb59a1

sumerjabri (approved):

PR #3583, opened 2024-08-01
Validate create folder API

phuongnq:
https://github.com/craftersoftware/craftercms/issues/967

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
@@ -317,6 +317,7 @@ public class ContentServiceImpl implements ContentService, ApplicationContextAwa
     @Valid
     @ValidateAction(type = Type.CREATE)
     @HasPermission(type = DefaultPermission.class, action = PERMISSION_CONTENT_WRITE)
+    @RequireSiteExists
     public void writeContent(@SiteId String site,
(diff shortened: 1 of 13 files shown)


src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
```diff
@@ -681,6 +688,17 @@ public void notifyContentEvent(@ValidateStringParam String site, @ValidateSecure
         applicationContext.publishEvent(new ContentEvent(securityService.getAuthentication(), site, path));
     }
 
+    @Override
+    @RequireSiteExists
```
jmendeza, line 700: @phuongnq please consider these scenarios as well: https://github.com/craftercms/craftercms/issues/7047

Also, please take a look at the other API V1 endpoints and implement a similar approach as needed

jmendeza (changes requested):

src/main/java/org/craftercms/studio/impl/v1/service/content/ContentServiceImpl.java
phuongnq: Updated to support curly braces.
Also, update similar APIs to validate the writing path:

```
rename-folder
write-content
```

jmendeza (approved):

sumerjabri (approved):

PR #3595, opened 2024-08-15
Do not insert processed_commits record if the site does not exist

jmendeza:
Do not insert processed_commits record if the site does not exist (create site)
https://github.com/craftercms/craftercms/issues/7165


jmendeza:
This change has to do with this call during site creation:
https://github.com/craftercms/studio/blob/da9e16a1b3b357f7098e55a974dd48a4ecde003c/src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java#L523

sumerjabri (approved):

PR #3596, opened 2024-08-22
Add rate limit to the API /studio/api/2/users/forgot_password

phuongnq:
https://github.com/craftersoftware/craftercms/issues/985

### DIFF
--- a/src/main/java/org/craftercms/studio/controller/rest/v2/UsersController.java
+++ b/src/main/java/org/craftercms/studio/controller/rest/v2/UsersController.java
@@ -402,13 +402,17 @@ public class UsersController {
 
     @GetMapping(FORGOT_PASSWORD)
     public ResultOne<String> forgotPassword(@NotBlank @RequestParam(value = REQUEST_PARAM_USERNAME) String username) {
+        int delay = studioConfiguration.getProperty(SECURITY_SET_PASSWORD_DELAY, Integer.class);
         try {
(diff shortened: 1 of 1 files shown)


src/main/java/org/craftercms/studio/controller/rest/v2/UsersController.java
```diff
@@ -402,13 +402,17 @@ public Result getCurrentUserSsoLogoutUrl() {
...
+        int delay = studioConfiguration.getProperty(SECURITY_SET_PASSWORD_DELAY, Integer.class);
         try {
+            TimeUnit.SECONDS.sleep(delay);
             ValidationUtils.validateValue(new EsapiValidator(USERNAME), username, REQUEST_PARAM_USERNAME);
             userService.forgotPassword(username);
         } catch (ServiceLayerException e) {
             logger.error("Failed to process forgot password for user '{}'", username, e);
         } catch (ValidationException e) {
             logger.error("Validation error while processing forgot password for user '{}'", username, e);
+        } catch (InterruptedException e) {
```
jmendeza, line 414: This will cause the request to be ignored silently.
We should either return an error response or have a separate try-catch block to prevent the actual operation from being skipped.

jmendeza (changes requested):

jmendeza (approved):

sumerjabri (approved):

PR #3598, opened 2024-08-26
Adjust available actions when item is in system_processing state

jmendeza:
Adjust available actions when item is in system_processing state
https://github.com/craftercms/craftercms/issues/7096


jmendeza:
@sumerjabri we will need to review this once the publisher work is merged

jmendeza:
Superseeded by https://github.com/craftercms/studio/pull/3662

PR #3599, opened 2024-08-29
Modify BP policy to match with default validator

phuongnq:
https://github.com/craftersoftware/craftercms/issues/993

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
+++ b/src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
@@ -207,9 +207,7 @@ public class GitContentRepository implements ContentRepository, ServletContextAw
                             .replace(REPO_COMMIT_MESSAGE_USERNAME_VAR, username)
                             .replace(REPO_COMMIT_MESSAGE_PATH_VAR, path);
                     commitId = helper.commitFiles(repo, siteId, comment, user, path);
-                    if (commitId != null) {
-                        insertProcessedCommitId(siteId, commitId);
-                    }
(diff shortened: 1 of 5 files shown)


src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
```diff
@@ -207,7 +207,9 @@ public String writeContent(String siteId, String path, InputStream content) {
                             .replace(REPO_COMMIT_MESSAGE_USERNAME_VAR, username)
                             .replace(REPO_COMMIT_MESSAGE_PATH_VAR, path);
                     commitId = helper.commitFiles(repo, siteId, comment, user, path);
-                    insertProcessedCommitId(siteId, commitId);
+                    if (commitId != null) {
```
sumerjabri: This is not related to the ticket. Usually, we need a ticket associated with these updates.

jmendeza: @sumerjabri New ticket to take care of this part
https://github.com/craftersoftware/craftercms/issues/996

Euquimides (approved):
Tested the new `Project Policy Configuration` regex in different BPs. The regex is working fine against the validator.

src/main/java/org/craftercms/studio/impl/v1/repository/git/GitContentRepository.java
sumerjabri: Added PR https://github.com/craftercms/studio/pull/3600

sumerjabri (approved):

PR #3615, opened 2024-09-12
Make role names, group names, and permission names case-insensitive

phuongnq:
https://github.com/craftercms/craftercms/issues/7001


jmendeza (approved):

sumerjabri (approved):

sumerjabri (changes requested):
Let's create two new classes:
- `Role`
- `Group`

These classes have a compare overload that compares things case-insensitively.

This way we never care if we forget to have a .toLowercase() in the future.

Refactor the code to use these new objects instead of strings.

phuongnq:
@sumerjabri I updated to add new `record` types `NormalizedGroup` and `NormalizedRole`. These types have the group/role name in lowercase and we can reuse the `equals` of `record` which compares if `name` has the same value. The method `toString()` is overridden to return the string value of the name.

sumerjabri (approved):

src/main/java/org/craftercms/studio/impl/v1/service/content/ContentTypeServiceImpl.java
```diff
@@ -186,9 +188,13 @@ public List<ContentTypeConfigTO> getAllowedContentTypesForPath(@ValidateStringPa
         }
     }
 
-    protected void addContentTypes(String site, Set<String> userRoles, ContentTypeConfigTO config,
+    protected void addContentTypes(String site, Set<NormalizedRole> userRoles, ContentTypeConfigTO config,
                                    List<ContentTypeConfigTO> contentTypes) {
-        boolean isAllowed = this.isUserAllowed(userRoles, config);
+        boolean isAllowed = this.isUserAllowed(userRoles
```
jmendeza: Let's change `isUserAllowed` signature to accept a Set of the new normalized object

src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
```diff
@@ -158,7 +160,7 @@ public Set<String> getUserPermissions(@ValidateStringParam final String site,
         if (StringUtils.isNotEmpty(site)) {
```
jmendeza, line 151: Please see if we can remove `groups` from this method. It seems it is always null except for a call from `get-user-permissions.get.groovy` (remove this API ??)
cc @sumerjabri

src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
```diff
@@ -170,7 +172,11 @@ public Set<String> getUserPermissions(@ValidateStringParam final String site,
             if (path.indexOf("/site") == 0) { // If it's content a file
                 try {
                     ContentTypeConfigTO config = contentTypeService.getContentTypeForContent(site, path);
-                    boolean isAllowed = contentTypeService.isUserAllowed(roles, config);
+                    boolean isAllowed = contentTypeService.isUserAllowed(roles
```
jmendeza: Change `isUserAllowed` to accept a Set of the new normalized object

src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
```diff
@@ -603,8 +610,8 @@ public boolean isSystemAdmin(@ValidateStringParam String username) {
 
         boolean toRet = false;
         if (CollectionUtils.isNotEmpty(roles)) {
-            for (String role : roles) {
-                if (StringUtils.equalsIgnoreCase(role, SYSTEM_ADMIN_ROLE)) {
+            for (NormalizedRole role : roles) {
+                if (role.equals(new NormalizedRole(SYSTEM_ADMIN_ROLE))) {
```
jmendeza: These are going to be created frequently, let's add constants for `new NormalizedRole(SYSTEM_ADMIN_ROLE)` and `new NormalizedRole(ADMIN_ROLE)`

src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
```diff
@@ -626,13 +633,12 @@ public boolean isSiteAdmin(@ValidateStringParam String username, String site) {
...
-                Map<String, List<String>> roleMappings = configurationService.getRoleMappings(site);
+                Map<NormalizedGroup, List<NormalizedRole>> roleMappings = configurationService.getRoleMappings(site);
 
                 if (MapUtils.isNotEmpty(roleMappings)) {
                     for (Group group : groups) {
-                        String groupName = group.getGroupName();
-                        List<String> roles = roleMappings.get(groupName);
-                        if (roles != null && roles.contains(ADMIN_ROLE)) {
+                        List<NormalizedRole> roles = roleMappings.get(new NormalizedGroup(group.getGroupName()));
+                        if (roles != null && roles.contains(new NormalizedRole(ADMIN_ROLE))) {
```
jmendeza: These are going to be created frequently, let's add constants for `new NormalizedRole(SYSTEM_ADMIN_ROLE)` and `new NormalizedRole(ADMIN_ROLE)`

src/main/java/org/craftercms/studio/api/v2/dal/security/NormalizedGroup.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+
+import static java.util.Objects.requireNonNullElse;
+
+/**
+ * Normalized group to deal with case-insensitive while working with roles/permissions mappings
+ * @param name name of the group
+ */
+public record NormalizedGroup(String name) {
```
jmendeza, line 27: `NormalizedGroup` and `NormalizedRole` seem identical. Any reason why we would want to keep both?

jmendeza (changes requested):

src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
sumerjabri: Yes, please remove this API, `api/1/security/get-user-permissions.json`.

src/main/java/org/craftercms/studio/impl/v1/service/content/ContentTypeServiceImpl.java
phuongnq: Updated

src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
phuongnq: Removed the API `api/1/security/get-user-permissions.json` and updated to remove `groups` from the method.

src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
phuongnq: Updated

src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
phuongnq: Updated

src/main/java/org/craftercms/studio/impl/v1/service/security/SecurityServiceImpl.java
phuongnq: Updated

src/main/java/org/craftercms/studio/api/v2/dal/security/NormalizedGroup.java
phuongnq: @jmendeza 
In the code, we have many data structures such as:
```
Map<String, Map<NormalizedRole, List<Node>>>
Map<NormalizedRole, List<Node>> siteRoles
Map<NormalizedGroup, List<NormalizedRole>> rolesMap

```
We can have a single class such as `NormalizedString` but it would make the code less readable. And we won't know what type of data the Map key/value is.

jmendeza: Got it.

jmendeza (approved):

PR #3618, opened 2024-09-17
Add maxKeys to S3 list API

phuongnq:
https://github.com/craftercms/craftercms/issues/6799


src/main/java/org/craftercms/studio/impl/v2/service/aws/s3/AwsS3ServiceImpl.java
```diff
@@ -147,22 +148,37 @@ public List<S3Item> listItems(@SiteId String siteId,
...
                     .map(p -> {
                         String relativeKey = StringUtils.removeStart(p.prefix(), profile.getPrefix());
                         return new S3Item(StringUtils.removeEnd(relativeKey, delimiter), relativeKey, true, profile.getBucketName(), profile.getPrefix());
                     })
                     .forEach(items::add);
+            commonPrefixesCount += page.commonPrefixes().size();
 
-            page.contents().stream()
+            // Do not fetch content key if it exceeded the maxKeys but continue to fetch prefixes
+            if (items.size() >= maxKeys + commonPrefixesCount) {
```
jmendeza: Why not have a counter for the contents (which are the ones being limited) instead of prefixes?
So condition would be like this:
```
contentsCount >= maxKeys
```

jmendeza (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/aws/s3/AwsS3ServiceImpl.java
phuongnq: Refactored the method `listItems` a bit and use `contentsCount`

jmendeza (approved):

sumerjabri (approved):

PR #3626, opened 2024-10-04
Unlocking a locked item should throw a warning in the logs rather than an error

phuongnq:
https://github.com/craftercms/craftercms/issues/7355


sumerjabri (changes requested):
Should we remove the exception `ContentAlreadyUnlockedException` if it's not going to be used?

phuongnq:
@sumerjabri I removed the unsued class `ContentAlreadyUnlockedException`

jmendeza (approved):

sumerjabri (approved):

PR #3629, opened 2024-10-11
Use tryLock to sync with repo in the case of watcher

phuongnq:
https://github.com/craftercms/craftercms/issues/7319


phuongnq:
close in favor of a better solution to make sure the deadlock does not happen

PR #3630, opened 2024-10-14
Lock with multiple lock keys with the order of the parameters

phuongnq:
https://github.com/craftercms/craftercms/issues/7319

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v2/utils/StudioUtils.java
+++ b/src/main/java/org/craftercms/studio/api/v2/utils/StudioUtils.java
@@ -18,6 +18,7 @@ package org.craftercms.studio.api.v2.utils;
 
 import org.craftercms.commons.http.RequestContext;
 import org.craftercms.studio.api.v1.constant.StudioConstants;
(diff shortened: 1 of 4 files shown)


jmendeza (changes requested):
We use the StudioUtils.getSyncFromRepoLockKey in the SyncFromRepositoryTask and DeploymentServiceImpl.
I think what we should do is make sure that we lock sandbox repo whenever we lock sync-from-repo, in the same action.
Let's make StudioUtils.getSyncFromRepoLockKey private. And have a couple helper methods to lock/unlock both keys in the right order.

lockSandboxSync(lockService, siteId)
unlockSandboxSync(lockService, siteId)
Then update DeploymentServiceImpl to call those instead of locking site sync only (the git operations will also lock on sandbox repo, but that will not be an issue since it will be the same thread).
So now no code path can lock sync first and sandbox repo later.

phuongnq:
I updated per discussed

jmendeza (approved):

sumerjabri (approved):

PR #3636, opened 2024-10-22
Sync from repo updates

jmendeza:
https://github.com/craftersoftware/craftercms/issues/1030
Reduce dependency related statements in generated script
Partition into chunks the list of changes to ingest
Time measurement updates
Add sync lock key back reduce thread inter locking
Add index on dependency.target_path

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/utils/TimeUtils.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/utils/TimeUtils.java
@@ -18,62 +18,110 @@ package org.craftercms.studio.impl.v2.utils;
 
 import org.craftercms.studio.api.v2.utils.function.ThrowingRunnable;
 import org.slf4j.Logger;
+import org.slf4j.event.Level;
+import org.slf4j.spi.LoggingEventBuilder;
 import org.springframework.util.function.ThrowingSupplier;
 
 public class TimeUtils {
     /**
-     * High order function to inject execution time if the logger is enabled trace.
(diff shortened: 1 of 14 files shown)


src/main/java/org/craftercms/studio/impl/v2/utils/TimeUtils.java
```diff
@@ -18,64 +18,112 @@
...
      * @param message      a message to flag the execution method
+     * @param level        the log level
      * @param <T>          generic type of the method return value
      * @return the result of the method
      */
-    public static <T> T logExecutionTimeThrowing(ThrowingSupplier<T> supplier, Logger methodLogger, String message) throws Exception {
+    public static <T> T logExecutionTimeThrowing(ThrowingSupplier<T> supplier, Logger methodLogger, String message, Level level) throws Exception {
         long startTime = 0;
-        if (methodLogger.isTraceEnabled()) {
+        LoggingEventBuilder loggingEventBuilder = methodLogger.makeLoggingEventBuilder(level);
```
sumerjabri: Can we avoid this method call if we're not at the requested log level?

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v2/utils/TimeUtils.java
jmendeza: Updated

sumerjabri (approved):

PR #3651, opened 2024-11-07
Fix API /content/history is very slow with big sites

phuongnq:
https://github.com/craftersoftware/craftercms/issues/1052

### DIFF
--- a/src/main/resources/crafter/studio/database/upgrade/4.2.x/4.2.0.15-to-4.2.0.16.sql
+++ b/src/main/resources/crafter/studio/database/upgrade/4.2.x/4.2.0.15-to-4.2.0.16.sql
@@ -1,0 +1,20 @@
+/*
+ * Copyright (C) 2007-2024 Crafter Software Corporation. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as published by
+ * the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+ALTER TABLE `audit`
+    ADD INDEX `audit_primary_target_value_idx` (`primary_target_value`) ;
+
+UPDATE `_meta` SET `version` = '4.2.0.16' ;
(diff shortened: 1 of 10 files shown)


src/main/resources/crafter/studio/database/upgrade/4.2.x/4.2.0.15-to-4.2.0.16.sql
```diff
@@ -0,0 +1,20 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+ALTER TABLE `audit`
+    ADD INDEX `audit_primary_target_value_idx` (`primary_target_value`) ;
```
jmendeza, line 18: @phuongnq have we tested the impact of this index in the insertion performance?

src/main/resources/org/craftercms/studio/api/v2/dal/AuditDAO.xml
```diff
@@ -280,6 +285,18 @@
         AND origin = 'API'
     </select>
 
+    <select id="getCommitAuthors" parameterType="java.util.Map" resultMap="CommitAuthorMap">
+        SELECT a.commit_id, u.*
+        FROM audit a
+        INNER JOIN user u on u.username = a.actor_id
+        WHERE a.commit_id IN
```
jmendeza: Should we filter by site id? Can we have the same commit id in different sites if cloning from the same remote?

jmendeza (changes requested):

src/main/resources/org/craftercms/studio/api/v2/dal/AuditDAO.xml
phuongnq: @jmendeza It's best to filter with the site id so I updated to include this

src/main/resources/crafter/studio/database/upgrade/4.2.x/4.2.0.15-to-4.2.0.16.sql
phuongnq: @jmendeza For a site with `470413` record in `audit` table, it took around 2.1s for a write API round-trip to the client (which includes the insert to audit log)
May need the QA team to double-test the impact on a big site with a larger audit table cc @sumerjabri

phuongnq: A direct INSERT into `audit` with the same table size of `470413` is instant (~0.007s)
Screenshot from 2024-11-08 14-11-43

jmendeza (approved):

sumerjabri (approved):

PR #3657, opened 2024-11-18
Add an audit log when an Encryption Token is created

phuongnq:
https://github.com/craftersoftware/craftercms/issues/1062

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/service/security/internal/EncryptionServiceInternalImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/service/security/internal/EncryptionServiceInternalImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2007-2022 Crafter Software Corporation. All Rights Reserved.
+ * Copyright (C) 2007-2024 Crafter Software Corporation. All Rights Reserved.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 3 as published by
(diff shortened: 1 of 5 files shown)


src/main/java/org/craftercms/studio/impl/v2/service/security/internal/EncryptionServiceInternalImpl.java
```diff
@@ -19,51 +19,90 @@
...
         }
         try {
             Thread.sleep(delay * 1000);
-            return textEncryptor.encrypt(text);
+            String encryptedToken = textEncryptor.encrypt(text);
+
+            var auth = SecurityContextHolder.getContext().getAuthentication();
+            String resolvedSiteId = StringUtils.isNotEmpty(siteId) ? siteId
+                    : studioConfiguration.getProperty(CONFIGURATION_GLOBAL_SYSTEM_SITE);
+            String targetId = UUID.randomUUID().toString();
```
sumerjabri: Can we store an MD5 sum of the encrypted token? That will let us connect the final token to the user and the audit record. You can always MD5 sum the token and match it to the audit record.

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/security/internal/EncryptionServiceInternalImpl.java
```diff
@@ -19,51 +19,90 @@
...
-    public void setTextEncryptor(TextEncryptor textEncryptor) {
-        this.textEncryptor = textEncryptor;
+    /**
+     * Create encryption audit log
+     * @param siteId site identifier
+     * @param actor actor
+     * @param targetId encryption target id
+     */
+    private void createEncryptionAuditLog(String siteId, String actor, String targetId) {
+        SiteFeed site = siteFeedMapper.getSite(Map.of(SITE_ID, siteId));
```
jmendeza: Let's use SiteDAO instead of SiteFeedMapper.

jmendeza (changes requested):

src/main/java/org/craftercms/studio/impl/v2/service/security/internal/EncryptionServiceInternalImpl.java
phuongnq: Updated to use MD5 sum

src/main/java/org/craftercms/studio/impl/v2/service/security/internal/EncryptionServiceInternalImpl.java
phuongnq: Updated to use SiteDAO

jmendeza (approved):

sumerjabri (approved):

PR #3662, opened 2024-11-27
Adjust available actions when item is in system_processing state

jmendeza:
https://github.com/craftercms/craftercms/issues/7096
Prevent clipboard operations on items in system_processing state
Adjust available actions when item is in system_processing state

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/service/publish/internal/PublishServiceInternalImpl.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/service/publish/internal/PublishServiceInternalImpl.java
@@ -601,7 +601,7 @@ public class PublishServiceInternalImpl implements PublishService, ApplicationCo
             }
             return publishPackage.getId();
         } catch (IOException e) {
-            logger.error("Failed to submit publish package: ", e);
+            logger.error("Failed to submit publish package", e);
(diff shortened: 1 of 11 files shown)


src/main/java/org/craftercms/studio/impl/v2/service/publish/internal/PublishServiceInternalImpl.java
```diff
@@ -610,6 +606,34 @@ protected long buildPublishPackage(final Site site,
...
+                                                Collection<PublishItem> publishItems) throws AuthenticationException, ServiceLayerException {
+        Collection<String> allPaths = null;
+        boolean clearSystemProcessing = false;
+        try {
+            allPaths = publishItems.stream()
+                    .map(PublishItem::getPath)
+                    .collect(toSet());
+
+            if (itemServiceInternal.isSystemProcessing(site.getSiteId(), allPaths)) {
+                throw new ServiceLayerException("Failed to submit publish package: Some items are being processed by the system");
```
sumerjabri, line 620: Why is there a `:` in the sentence?

sumerjabri (changes requested):
Please fix the exception message.

sumerjabri (approved):

PR #3680, opened 2024-12-23
Publisher API fixes

jmendeza:
Fix approve/reject APIs issue where it would expect additional packageIds params
Fix empty path list query when trying to publish directories with no changes


src/main/java/org/craftercms/studio/api/v2/dal/publish/PublishPath.java
```diff
@@ -0,0 +1,41 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.craftercms.studio.api.v2.dal.publish;
+
+/**
+ * Simplified Publish item path-action pair.
+ */
+public class PublishPath {
```
jmendeza: Rename to PathActionPair

sumerjabri (approved):

PR #3682, opened 2025-01-07
Retire CMIS APIs entirely

phuongnq:
https://github.com/craftercms/craftercms/issues/7513


sumerjabri (approved):

jmendeza (changes requested):
Is this related as well?
https://github.com/craftercms/studio/blob/develop/src/main/webapp/default-site/scripts/pages/browseCMIS.groovy

phuongnq:
It seems the page `browseCMIS.groovy` is no longer in use so I have removed it!

jmendeza (approved):

PR #3683, opened 2025-01-07
Check if path regex is empty for update_item_states_by_query API

phuongnq:
https://github.com/craftercms/craftercms/issues/7521


src/main/java/org/craftercms/studio/controller/rest/v2/WorkflowController.java
```diff
@@ -124,7 +127,8 @@ public Result updateItemStatesByQuery(@Valid @RequestBody UpdateItemStatesByQuer
             throws SiteNotFoundException {
         UpdateItemStatesByQueryRequestBody.Query query = requestBody.getQuery();
         ItemStatesUpdate update = requestBody.getUpdate();
-        workflowService.updateItemStatesByQuery(query.getSiteId(), query.getPath(),
+        String resolvedPathRegex = StringUtils.isNotEmpty(query.getPath()) ? query.getPath() : DEFAULT_QUERY_BY_PATH_REGEX;
```
jmendeza, line 129: This is failing when checking permissions. Does it make sense to check permission for a regex path?
Let's remove the `@ProtectedResourceId` in [updateItemsStatesByQuery](https://github.com/craftercms/studio/blob/develop/src/main/java/org/craftercms/studio/impl/v2/service/workflow/WorkflowServiceImpl.java#L77) and continue to pass null as the regex. That way the query gets simplified in those scenarios instead of trying to match `.*` for all paths

jmendeza (changes requested):

src/main/java/org/craftercms/studio/controller/rest/v2/WorkflowController.java
phuongnq: Hi @jmendeza If we decide to remove the `@ProtectedResourceId`, does this means a user needs the permission `set_item_states` on the default regex `.*` when performing an update for item states even if the regex is not the ALL path one?
For example, if a developer does not have `set_item_states` permission on the regex `.*` but with the regex `/site/website/.*`, should we mark this as failed because of the lack of permission?

```
<rule regex="/site/website/.*">
            <allowed-permissions>
                <permission>set_item_states</permission>
            </allowed-permissions>
        </rule>
```

Screenshot from 2025-01-08 13-44-59

Just bring this for discussion before we decide to make user to have the permission on ALL path to perform update by query.

The permission checking kind of 'works' for the case regex. For example, if a user updates the state with the regex `/site/website/.*` and in the permission mapping we have this regex or the ALL path `.*`, this will be a match.

phuongnq: Other than this, I updated also to check if the path is a valid regex (See https://github.com/craftercms/studio/pull/3683/commits/fed9adbf54be087b9b9ff5080734a08117d02db4). We have performed the same for the API `GET item_states` but let me know if we have other opinions on this check!

jmendeza: Thanks @phuongnq . I see. I think I'm fine with it since it does not change the API behavior.

jmendeza (approved):

sumerjabri (approved):

PR #3691, opened 2025-01-17
Throw exception with write-content API when committer identity unknown

phuongnq:
https://github.com/craftercms/craftercms/issues/7599


src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
```diff
@@ -1281,6 +1286,9 @@ public String commitFiles(Repository repo, String site, String comment, PersonId
                 restorePaths(repo, site, paths);
                 logger.error("Failed to commit files to git in site '{}' paths '{}'", site,
                              ArrayUtils.toString(paths), e);
+				if (cause instanceof GitCliOutputException && (cause.getMessage().toLowerCase().contains(GIT_COMMITTER_IDENTITY_UNKNOWN_MESSAGE))) {
```
jmendeza: Can we add an [exception resolver](https://github.com/craftercms/studio/blob/develop/src/main/java/org/craftercms/studio/impl/v2/utils/git/cli/PatternFindingGitCliExceptionResolver.java) implementation for this?

jmendeza (changes requested):

src/main/java/org/craftercms/studio/api/v2/utils/GitRepositoryHelper.java
phuongnq: Added `CommitterIdentityUnknownExceptionResolver`

jmendeza (approved):

sumerjabri (approved):

PR #3708, opened 2025-02-04
Implement externally managed flag for user groups

phuongnq:
https://github.com/craftersoftware/craftercms/issues/1080

### DIFF
--- a/src/main/java/org/craftercms/studio/api/v2/dal/GroupDAO.java
+++ b/src/main/java/org/craftercms/studio/api/v2/dal/GroupDAO.java
@@ -53,11 +53,11 @@ public interface GroupDAO {
      * @param orgId organization id
      * @param groupName group name
      * @param groupDescription  group description
-     * @param externallyManaged 1 if group is externally managed, 0 otherwise
+     * @param externallyManaged true if group is externally managed, false otherwise
      * @return Number of affected rows in DB
      */
     Integer createGroup(@Param(ORG_ID) long orgId, @Param(GROUP_NAME) String groupName,
-                        @Param(GROUP_DESCRIPTION) String groupDescription, @Param(EXTERNALLY_MANAGED) int externallyManaged);
+                        @Param(GROUP_DESCRIPTION) String groupDescription, @Param(EXTERNALLY_MANAGED) boolean externallyManaged);
 
     /**
      * Update group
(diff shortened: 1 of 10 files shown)


src/main/java/org/craftercms/studio/api/v2/dal/GroupDAO.java
```diff
@@ -132,9 +132,11 @@ List<User> getGroupMembers(@Param(GROUP_ID) long groupId, @Param(OFFSET) int off
      *
      * @param groupId group identifier
      * @param userIds list of user identifiers
+	 * @param externallyManaged 1 if group is externally managed, 0 otherwise
      * @return Number of rows affected in DB
      */
-    Integer addGroupMembers(@Param(GROUP_ID) long groupId, @Param(USER_IDS) List<Long> userIds);
+    Integer addGroupMembers(@Param(GROUP_ID) long groupId, @Param(USER_IDS) List<Long> userIds,
```
jmendeza, line 138: Will mybatis handle a boolean here as the java type?

jmendeza (changes requested):

src/main/java/org/craftercms/studio/api/v2/dal/GroupDAO.java
phuongnq: It is not a boolean. In mybatis, we perform the following query:
```
<insert id="addGroupMembers" parameterType="java.util.Map" >
        INSERT IGNORE INTO `group_user` (user_id, group_id, externally_managed) VALUES
        <foreach collection="userIds" item="userId" index="index" separator="," >
            (#{userId}, #{groupId}, #{externallyManaged})
        </foreach>
    </insert>
```

This aligns with how we are doing with mybatis `createGroup`

jmendeza: It is logically a boolean. If this param is a boolean then we don't need to convert to int (MariaDB will do it automatically). So we wouldn't need to do something like this anymore: `externallyManaged ? 1 : 0` 

I think we should change `createGroup` as well.

phuongnq: Sound good. I updated to use `boolean` instead of `int`

jmendeza (approved):

sumerjabri (approved):

PR #3725, opened 2025-02-12
Return item metadata in calculate and recalculate publish package APIs

jmendeza:
Return item metadata in calculate and recalculate publish package APIs


sumerjabri (changes requested):
Please update the API spec

jmendeza:
> Please update the API spec

Done

sumerjabri (approved):

PR #3734, opened 2025-02-18
Replace generated SQL script with Mybatis

phuongnq:
https://github.com/craftersoftware/craftercms/issues/1015


src/main/java/org/craftercms/studio/api/v2/dal/ItemDAO.java
```diff
@@ -673,4 +676,47 @@ boolean matchItemState(@Param(SITE_ID) String siteId,
 	 * @param paths  the paths to update
 	 */
 	void updateParentId(@Param(SITE_ID) long siteId, @Param(PATHS) Collection<String> paths);
+
+	/**
+	 * Update a new page children.
+	 * This should be called when a new page (index.xml) is created in an already existing folder
```
jmendeza: Shouldn't it rather be called when the page is `deleted` but the folder and the children remain?

src/main/java/org/craftercms/studio/impl/v2/sync/SyncFromRepositoryTask.java
```diff
@@ -523,27 +498,30 @@ private User getRepoOperationUser(String operationAuthor, Map<String, User> cach
...
-		for (RepoOperation repoOperation : repoOperations) {
-			User user = getRepoOperationUser(repoOperation.getAuthor(), cachedUsers);
-			switch (repoOperation.getAction()) {
-				case CREATE, COPY -> processCreate(site, repoOperation, user, repoOperationsScriptPath, allAncestors);
-				case UPDATE -> processUpdate(site, repoOperation, user, repoOperationsScriptPath);
-				case DELETE -> processDelete(site, repoOperation, repoOperationsScriptPath);
-				case MOVE -> processMove(site, repoOperation, user, repoOperationsScriptPath, allAncestors);
-				default -> logger.error("Failed to process unknown repo operation '{}' in site '{}'",
-					site.getSiteId(), repoOperation.getAction());
+		for (List<RepoOperation> batchRepoOperations : ListUtils.partition(repoOperations, DalUtils.MY_BATIS_QUERY_BATCH_SIZE)) {
```
jmendeza: Should we have a single `sqlSession` outside of the loops? And flush periodically if needed

src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
```diff
@@ -446,46 +465,27 @@ private void processCreatedFiles(String siteId, String creator,
...
-
-			MutableLong itemCount = new MutableLong(0);
-			ThrowingRunnable checkCounter = getCheckCounterFunction(itemCount, siteId, studioDBScriptRunner, scriptPath);
+		MutableLong itemCount = new MutableLong(0);
+		try (SqlSession sqlSession = sqlSessionFactory.openSession(ExecutorType.BATCH)) {
+			ThrowingRunnable checkCounter = getCheckCounterFunction(sqlSession, itemCount, siteId);
 			contentRepositoryV2.forAllSitePaths(siteId,
 				directory -> {
-					processCreatedDirectory(scriptPath, site.getId(), directory, userObj.getId(), now);
+					processCreatedDirectory(sqlSession, site.getSiteId(), directory, userObj.getId(), now);
```
jmendeza: Can we get the mapper out of the session and pass it here instead of the actual sqlSession object? That would create a single mapper for all updates.

src/main/java/org/craftercms/studio/impl/v2/utils/DependencyUtils.java
```diff
@@ -43,54 +41,45 @@ public class DependencyUtils {
...
 	 * @param revalidate        if true, the existing dependencies pointing to the path will be set to valid=true
-	 * @throws IOException if an error occurs while updating the script
 	 */
-	public static void addDependenciesScriptSnippets(String siteId, String path, String oldPath,
-							 Path file, DependencyService dependencyService,
-							 boolean cleanExisting, boolean revalidate)
-		throws IOException {
+	public static void updateDependencies(String siteId, String path, String oldPath,
+										  DependencyService dependencyService, SqlSession sqlSession, boolean cleanExisting, boolean revalidate) {
+		DependencyDAO dependencyDao = sqlSession.getMapper(DependencyDAO.class);
```
jmendeza: Can we get the mapper in the caller just once? Here it seems we will get multiple `DependencyDAO` mappers, one per updated path.

src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
```diff
@@ -165,6 +161,8 @@ public class SiteServiceImpl implements SiteService, ApplicationContextAware {
 
 	protected UserDAO userDao;
 
+	SqlSessionFactory sqlSessionFactory;
```
jmendeza: Should this be private or protected ?

jmendeza (changes requested):

src/main/java/org/craftercms/studio/api/v2/dal/ItemDAO.java
phuongnq: Updated

src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
phuongnq: Updated to use `protected` to align with this class implementation.

src/main/java/org/craftercms/studio/impl/v2/utils/DependencyUtils.java
phuongnq: I updated to send `dependencyDao` and `itemDao` from the caller.

src/main/java/org/craftercms/studio/impl/v2/sync/SyncFromRepositoryTask.java
phuongnq: Updated

src/main/java/org/craftercms/studio/impl/v1/service/site/SiteServiceImpl.java
phuongnq: I updated to send `dependencyDao` and `itemDao` from the caller. 
Please note that we cannot use the initialized DAO beans as it will update DB immediately instead of using the session.

jmendeza (approved):

sumerjabri (approved):

PR #3735, opened 2025-02-18
API v2 Refactor

jmendeza:
API v2 Refactor
https://github.com/craftercms/craftercms/issues/7542


src/main/api/studio-api.yaml
```diff
@@ -9369,17 +9369,13 @@ components:
             type: object
             properties:
                 item:
-                    $ref: '#/components/schemas/ContentItem'
+                    $ref: '#/components/schemas/ContentItemV1'
```
sumerjabri: What's this?

jmendeza: That API (/api/2/configuration/get_configuration_history) spec was not matching its actual response.

sumerjabri: Will we have two return types, ContentItem and ContentItemV1?

jmendeza: I will update `get_configuration_history` to return the new one. 
Notice ContentItemV1 is still returned by some V1 APIs. So it can be removed once those APIs are retired:
- /api/1/services/api/1/content/get-item.json
- /api/1/services/api/1/content/get-item-orders.json
- /api/1/services/api/1/content/get-pages.json
- /api/1/services/api/1/content/get-items-tree.json

jmendeza:
Notice that because of the API breaking changes, some UI features will be affected by this PR.
@sumerjabri @rart

sumerjabri (approved):

PR #3758, opened 2025-04-21
Redirect command output to files to prevent process locking when large outpu

jmendeza:
Redirect command output to files to prevent process locking when large outputs
https://github.com/craftercms/craftercms/issues/7939

## Summary by CodeRabbit

- **Refactor**
  - Enhanced Git command execution by improving output handling with temporary files and configurable output size limits.
  - Unified repository directory references by using directory objects instead of string paths across Git operations.
  - Added configuration to control maximum bytes read from Git process output for better resource management.

### DIFF
--- a/src/main/java/org/craftercms/studio/impl/v2/utils/git/GitCli.java
+++ b/src/main/java/org/craftercms/studio/impl/v2/utils/git/GitCli.java
@@ -16,7 +16,7 @@
 package org.craftercms.studio.impl.v2.utils.git;
 
 import org.apache.commons.collections4.ListUtils;
-import org.apache.commons.io.IOUtils;
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.craftercms.studio.api.v2.exception.git.cli.GitCliException;
@@ -34,8 +34,6 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStream;
-import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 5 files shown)


src/main/java/org/craftercms/studio/impl/v2/utils/git/GitCli.java
```diff
@@ -175,10 +176,10 @@ private void handleErrorExitValue(String directory, GitCliOutputExceptionResolve
 			.orElse(new GitCliOutputException(exitValue, errorMessage));
 	}
 
-	private void handleProcessTimeout(Process p, String directory, InputStream processInputStream, InputStream processErrorStream) throws IOException {
+	private void handleProcessTimeout(Process p, String directory, File stdOutFile, File stdErrFile) throws IOException {
 		// Read available bytes, avoiding blocking
-		String stdOutput = new String(processInputStream.readNBytes(processInputStream.available()));
-		String errorOutput = new String(processErrorStream.readNBytes(processErrorStream.available()));
+		String stdOutput = Files.readString(stdOutFile.toPath());
```
sumerjabri: This can read until it exhausts all memory. Can we read the first x bytes for logging purposes and stop? Should we just log the path where the output file exists for the admins to read the actual temp file and see the logs? Any other path other than reading an unknown-size file into RAM?

src/main/java/org/craftercms/studio/impl/v2/utils/git/GitCli.java
```diff
@@ -161,10 +162,10 @@ private String doExecuteGitCommand(final String directory, final GitCommandLine
 	}
 
 	private void handleErrorExitValue(String directory, GitCliOutputExceptionResolver exceptionResolver,
-					  Process p, InputStream processInputStream) throws IOException {
+									  Process p, File stdOutFile, File stdErrFile) throws IOException {
 		int exitValue = p.exitValue();
-		String errorOutput = IOUtils.toString(p.getErrorStream(), Charset.defaultCharset());
-		String stdOutput = IOUtils.toString(processInputStream, Charset.defaultCharset());
+		String errorOutput = Files.readString(stdErrFile.toPath());
```
sumerjabri: See other comment.

sumerjabri (changes requested):

src/main/java/org/craftercms/studio/impl/v2/utils/git/GitCli.java
```diff
@@ -164,8 +180,8 @@ private String doExecuteGitCommand(final String directory, final GitCommandLine
 	private void handleErrorExitValue(String directory, GitCliOutputExceptionResolver exceptionResolver,
 									  Process p, File stdOutFile, File stdErrFile) throws IOException {
 		int exitValue = p.exitValue();
-		String errorOutput = Files.readString(stdErrFile.toPath());
-		String stdOutput = Files.readString(stdOutFile.toPath());
+		String errorOutput = readFileFirstBytes(stdErrFile);
```
sumerjabri, line 183: I think we can do something like:
if loglevel = debug (or trace), then read the file and log the messages.

sumerjabri (changes requested):

sumerjabri (approved):

PR #3786, opened 2025-08-04
Move and update API

jmendeza:
Move and update API
Policy validation on write operations
https://github.com/craftercms/craftercms/issues/8086

## Summary by CodeRabbit

* **New Features**
  * Added a new API endpoint to move content and update it in a single operation.
  * Enabled specifying new content during move operations via request body.
  * Introduced support for content update lifecycle during move actions.

* **Bug Fixes**
  * Improved validation annotations for content move and write operations.

* **Tests**
  * Added tests to verify entitlement validation during content write and delete operations.

* **Documentation**
  * Updated API documentation to include the new move-and-update endpoint.

### DIFF
--- a/src/main/api/studio-api.yaml
+++ b/src/main/api/studio-api.yaml
@@ -5046,6 +5046,66 @@ paths:
                 '500':
                     $ref: '#/components/responses/InternalServerError'
 
+    /api/2/content/{siteId}/move-and-update:
+        post:
+            tags:
+                - content
+            summary: Move and update content
+            description: 'Required permission "content_write", "content_delete"'
+            operationId: contentMoveAndUpdate
+            parameters:
+                -   name: siteId
+                    in: path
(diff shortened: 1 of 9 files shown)


src/main/api/studio-api.yaml
```diff
@@ -5046,6 +5046,66 @@ paths:
                 '500':
                     $ref: '#/components/responses/InternalServerError'
 
+    /api/2/content/{siteId}/move-and-update:
```
sumerjabri: Seems like we switched to `_` in v2 instead of `-`. Please have a look.

sumerjabri (changes requested):

src/main/api/studio-api.yaml
jmendeza: You're right. Updated

sumerjabri (approved):

PR #3801, opened 2025-08-28
Prevent clearing locked_by on edit. Fix column name

jmendeza:
https://github.com/craftercms/craftercms/issues/8273
Prevent dropping old `locked_by` value on item update. Fix column name
Remove references to old submitter property

## Summary by CodeRabbit

* **New Features**
  * Item responses now include lastModifiedBy, createdBy, and lockedBy fields for clearer user metadata.
* **Refactor**
  * Updated user metadata mapping for Items to improve consistency.
  * The submitter field has been removed from Item responses.
  * Builder/API surface no longer exposes previousPath-related setter.


jmendeza:
Outside diff range comments:
One of the issues was `locked_by` being inadvertently cleared on edit.
It can still be cleared if set to null by the service layer and would be a valid scenario

sumerjabri (approved):

PR #3815, opened 2025-09-22
Fixed code and config to support LockByKey

avasquez614:
https://github.com/craftercms/craftercms/issues/8349

### DIFF
--- a/src/main/resources/crafter/studio/extension/overlay-config.properties
+++ b/src/main/resources/crafter/studio/extension/overlay-config.properties
@@ -14,6 +14,11 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+# Interval between cleanup runs of locks by key (in seconds)
+crafter.core.locks.byKey.cleanupIntervalSecs=3600
+# Threshold for warning about idle locks by key (in milliseconds)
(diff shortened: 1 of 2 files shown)


src/main/resources/crafter/studio/extension/overlay-config.properties
```diff
@@ -14,6 +14,11 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+# Interval between cleanup runs of locks by key (in seconds)
+crafter.core.locks.byKey.cleanupIntervalSecs=3600
+# Threshold for warning about idle locks by key (in milliseconds)
+crafter.core.locks.byKey.warnIdleMillis=600000
```
sumerjabri, line 20: Why is this different from `core`?

sumerjabri (changes requested):

src/main/resources/crafter/studio/extension/overlay-config.properties
avasquez614: Studio has heavier operations, so it's possible that we could have operations holding a key for 10 mins or more. I don't expect the same in Engine (for Preview and Delivery). But if you think this shouldn't be the case I can remove the override.

sumerjabri (approved):

PR #3944, opened 2026-06-09
Prevent matching parent path when looking for collisions on content copy/move

jmendeza:
Prevent matching parent path when looking for collisions on content copy/move
https://github.com/craftercms/craftercms/issues/8672

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file duplication naming behavior to more reliably handle collision detection when files with copy modifiers already exist.


jmendeza:
@coderabbitai review

sumerjabri (approved):

